PR #11, opened 2018-01-26
Add new extension GL_EXT_control_flow_attributes.

johnkslang:
This extension is open for discussion, which can be held here in this pull request.

Basically, it adds control-flow hints (very strong suggestions).  E.g.,

```
        [[unroll]]               for (int x = 0; i < 8; ++i) { ... }
        [[dont_unroll]]          while(...) { ... }
        [[dependency_infinite]]  do { ... } while(...);
        [[dependency_length(4)]] for (int x = 0; i < 8; ++i) { ... }
        [[flatten]]              if (...) { ... } else { ... }
        [[dont_flatten]]         switch(...) { ... }
```

See content of the extension for more detail. There is an implementation under way for this in glslang.

### DIFF
--- a/extensions/ext/GL_EXT_control_flow_attributes.txt
+++ b/extensions/ext/GL_EXT_control_flow_attributes.txt
@@ -1,0 +1,184 @@
+Name
+
+    EXT_control_flow_attributes
+
+Name Strings
+
+    GL_EXT_control_flow_attributes
+
+Contact
+
+    John Kessenich (johnkessenich 'at' google.com), Google
+
+Contributors
+
+    Khronos members
+
+Notice
+
+    Copyright (c) 2018 Google, Inc.
+
+Status
+
+    Draft.
+
+Version
+
+    Last Modified Date: 25-Jan-2018
+    Revision: 2
+
+Number
+
+    TBD
+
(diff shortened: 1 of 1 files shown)


Jasper-Bekkers:
Can't we just use the "unroll / loop" and "flatten / branch" terminology?

gwihlidal:
Yes, it would be nice to use the existing terminology (avoids having to compare two documentation sets to determine equivalency). I'm definitely happy to see this extension, though!

Anteru:
Note that loop/unroll and flatten/branch is allowed -- it's an alias for dont_unroll/dont_flatte.

gwihlidal:
Ok, the aliases are very nice! Thank you.

johnkslang:
Right, the attributes include the union of the SPIR-V terminology and other existing terminology, resulting in aliases.

Error/warning behavior has also now been added to the extension.

johnkslang:
This is implemented now in https://github.com/KhronosGroup/glslang/pull/1232.

johnkslang:
The glslang implementation of this is complete and merged into glslang now.

I will merge this EXT into the repository, but continued comments are welcome.

PR #17, opened 2018-05-03
Add a new GL_EXT_subgroupuniform_qualifier.

sheredom:
Add a new GL_EXT_dynamically_uniform_attribute extension that introduces
a variable declaration, function parameter, and function return type
attribute [[dynamically_uniform]] to signify that the value is uniform
across the shader.

This maps to the SPIR-V `Uniform` decoration.

### DIFF
--- a/extensions/ext/GL_EXT_subgroupuniform_qualifier.txt
+++ b/extensions/ext/GL_EXT_subgroupuniform_qualifier.txt
@@ -1,0 +1,164 @@
+Name
+
+    EXT_subgroupuniform_qualifier
+
+Name Strings
+
+    GL_EXT_subgroupuniform_qualifier
+
+Contact
+
+    Neil Henning (neil.henning 'at' amd.com)
+
+Contributors
+
+    Jeff Bolz, NVIDIA Corporation
+    John Kessenich, Google
+    Nicolai Hähnle, AMD
+
+Status
+
+    Draft
+
+Version
+
+    Last Modified Date:         September 19, 2018
+    Revision:                   1
+
+Number
+
+    TBD
+
+Dependencies
+
+    This extension requires GL_KHR_vulkan_glsl
+
+Overview
+
+    This extension adds a "subgroupuniformEXT" type qualifier and constructor,
+    such that for each dynamic instance of an object qualified by a subgroup
+    uniform type or constructed with a subgroup uniform constructor, all active
+    invocations in a subgroup compute the same result value.
+
+New Procedures and Functions
+
+    None.
+
+New Tokens
+
+    None.
+
+Modifications to GL_KHR_vulkan_glsl
+
+    Add to the "Mapping to SPIR-V" section
+    
+    Mapping of subgroupuniformEXT type qualifier:
+
+      subgroupuniformEXT -> Uniform decoration on variables
+
+Modifications to the OpenGL Shading Language Specification, Version 4.50
+
+    Including the following line in a shader can be used to control the
+    language features described in this extension:
+
+        #extension GL_EXT_subgroupuniform_qualifier : <behavior>
+
+    where <behavior> is as specified in section 3.3
+
+    New preprocessor #defines are added to the OpenGL Shading Language:
+
+        #define GL_EXT_subgroupuniform_qualifier     1
+
+    Add to section 3.6 Keywords:
+
+        subgroupuniformEXT
+
+    Add a new section:
+      
+      "4.X subgroupuniformEXT qualifier"
+
+      The subgroupuniformEXT qualifier can be used to assert that for a dynamic
+      instance of a variable or expression all active invocations in a subgroup
+      compute the same result value. In a declaration, it is syntactically
+      treated as a qualifier. It can be applied to:
+
+        * variable declarations qualified as *in*
+        * global variable declarations with no storage qualifier
+        * local variable declarations with no storage qualifier
+        * function parameter declarations and function return types.
+
+      Any other use on a declaration results in a compile-time error.
+
+      The subgroupuniformEXT qualifier can also be used with constructor syntax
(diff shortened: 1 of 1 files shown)


johnkslang:
Looks good.

Are there propagation rules for expressions? If so, e.g., copy the set of operators doing so from the nonuniform specification?  Maybe in general, we want lots of parallels with that?

johnkslang:
> I assume constants are dynamically uniform (no brainer)

and uniform/buffer too.

> I don't propagate the decorations in the SPIR-V at present

I'm not sure I fully understand, so just talking a bit:

- the front-end (before SPIR-V generation) should do all the propagation work the extension requires
- the SPIR-V consumer should not have to do chains of data-flow analysis: the points where it matters should be decorated

sheredom:
> the SPIR-V consumer should not have to do chains of data-flow analysis: the points where it matters should be decorated

The case I'm talking about is something like:

```
OpDecorate %1 Uniform
%1 = OpVariable int
%2 = OpLoad int %1 // Should this be decorated with Uniform too?
%3 = OpIAdd int %2, %2 // And this?
```

At present I decorate the places where the compiler cannot know uniformity, not all the places a thing happens to be dynamically uniform.

sheredom:
> and uniform/buffer too.

I agree that uniform should be, well, uniform, but buffer?

You could easily have divergence in a thread where invocations write to elements of a buffer and no barriers which would result in an unknown uniformity result being read by each invocation.

johnkslang:
So, these propagate in the GLSL, but that result is then not reflected in the SPIR-V?

What about the points of use that benefit from knowing uniformity? If being parsimonious with decorations, I'd understand decorating there, at the end of the chain, rather than the beginning of the chain. That was the goal for nonuniform too.

sheredom:
> So, these propagate in the GLSL, but that result is then not reflected in the SPIR-V?

At present yup, I was trying to avoid decoration bloat by only tagging the 'roots' as uniform. I'll see how hard it would be to propagate uniformity everywhere and get some data though.

sheredom:
For your example above:

```
[[dynamically_uniform]] bool x = (invocationID & 1) != 0;
```

My current patch to glslang will produce a warning here because invocationID is not dynamically uniform and thus you are assigning a value of unknown uniformity to a variable you have said is dynamically uniform.

> I think you should file a SPIR-V issue and the SPIR-V group should discuss this, before we proceed further with a GLSL extension. We probably need a new decoration.

I'll file an internal issue yup!

sheredom:
We've just updated the SPIR-V specification to clarify what our intention was with the `Uniform` decoration, which now fixes @jeffbolznv's concerns above I believe. See the details https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_changes_from_version_1_3_revision_3

I think that unblocks this issue and we can proceed to discuss the merits of this exact proposal?

jeffbolznv:
Agreed, my technical objections at the SPIR-V level have been addressed, now it is primarily an HLL syntax issue.

johnkslang:
Is there any pending discussion?

Let me know when you are ready for merge.

sheredom:
This extension has been modified based on the discussions at the Khronos F2F to use the same mechanism as the GL_EXT_nonuniform_qualifier extension to specify something is subgroup uniform.

extensions/ext/GL_EXT_subgroupuniform_qualifier.txt
```diff
@@ -0,0 +1,158 @@
...
+          }
+
+      This constructor syntax takes a single argument of any type and returns
+      the value with the same type, qualified with subgroupuniformEXT.
+
+      Only some operations discussed in Chapter 5 (Operators and Expressions)
+      can be applied to subgroup uniform value(s) and still yield a result that
+      is subgroup uniform. The operations that do so are listed below. When a
+      subgroup uniform value is operated on with one of these operators
+      (regardless of whether any and other operands are subgroup uniform), the
```
jeffbolznv: This needs to be changed from how nonuniformEXT worked. For nonuniformEXT, if any operand is nonuniform then the result is nonuniform. For subgroupuniform, we would want to say that the result is subgroupuniform only if all operands are subgroupuniform.

I didn't review the list of operations below, but it might need to be different from nonuniform as well.

sheredom:
@jeffbolznv I've done the changes you requested, but because I amended the commit and force pushed I can't work out _how_ to get into your review to state I've done it.

It was just a copy pasta issue, I've made it now say the result is only subgroup uniform is all operands of an operator are subgroup uniform (which matches the implementation I've done).

nhaehnle:
What happens if not all dynamic instances of the expression actually compute the same value? Can we explicitly say somewhere that this is undefined behavior, possibly as an Issue?

extensions/ext/GL_EXT_subgroupuniform_qualifier.txt
```diff
@@ -0,0 +1,164 @@
...
+
+    None.
+
+Modifications to GL_KHR_vulkan_glsl
+
+    Add to the "Mapping to SPIR-V" section
+    
+    Mapping of subgroupuniformEXT type qualifier:
+
+      subgroupuniformEXT -> subgroupuniformEXT decoration on variables
```
johnkslang: Is that the right mapping? It is not to Uniform? At least the case seems incorrect, but I didn't see where the SPIR-V subgroupuniformEXT was introduced.

sheredom: Ah no doh, this was meant to be the Uniform decoration in SPIR-V but I copy+pasta fail. Lemme fix that quick.

johnkslang:
I realized too late this is missing the new entry in the index in the README. I'll add it.

PR #18, opened 2018-06-25
Add GL_EXT_samplerless_texture_functions

aejsmith:
This adds a new extension that allows passing texture objects directly to texture query and lookup functions that do not require a sampler, specifically:

 - texelFetch
 - texelFetchOffset
 - textureSize
 - textureQueryLevels
 - textureSamples


pdaniell-nv:
Would this be a good candidate for being a "GL_KHR" extension?

PR #22, opened 2018-07-21
Relax rules for *sampler* to allow shadows.

johnkslang:
Shadow sampling does not strictly require a *samplerShadow*.
Addresses Khronos internal issue https://gitlab.khronos.org/spirv/SPIR-V/issues/160.


johnkslang:
Khronos decided to generalize this to both directions of (not) checking for shadow compatibility.

PR #27, opened 2018-10-02
Adding GL_EXT_shader_atomic_int64 Spec

AaronHaganAMD:
Adding GLSL spec for GL_EXT_shader_atomic_int64

Issue https://github.com/KhronosGroup/Vulkan-Docs/issues/796

### DIFF
--- a/extensions/ext/GL_EXT_shader_atomic_int64.txt
+++ b/extensions/ext/GL_EXT_shader_atomic_int64.txt
@@ -1,0 +1,132 @@
+Name
+
+     EXT_shader_atomic_int64
+
+Name Strings
+
+     GL_EXT_shader_atomic_int64
+
+Contact
+
+     Aaron Hagan, AMD (aaron.hagan 'at' amd.com)
+
+Contributors
+
+     Pat Brown, NVIDIA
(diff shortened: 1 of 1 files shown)


pdaniell-nv:
We touched on this idea briefly in the OpenGL/ES meeting today. Conceivably this extension in its current form could be put in the OpenGL registry and OpenGL implementation could implement it and expose the "GL_EXT_shader_atomic_int64" extension. Any reason not to?

drakos-amd:
I don't think there's a technical reason, but this is blocking the completion of the release of the corresponding Vulkan extension, so I'd suggest accepting this PR as is and making a decision on whether to turn this into a full GL extension separately.

pdaniell-nv:
Yeah, that's fine. I don't want to block this getting in.

johnkslang:
This can effectively be considered as in, being an EXT, I just leave it as a PR for about a week to trigger comments.

elviras9t:
I suppose, need also to solve coverage with GL_NV_shader_atomic_int64
(https://www.khronos.org/registry/OpenGL/extensions/NV/NV_shader_atomic_int64.txt)

PR #40, opened 2018-11-16
Add GL_EXT_scalar_block_layout to registry

Tobski:
Add GLSL specification component for https://github.com/KhronosGroup/Vulkan-Docs/issues/854


Tobski:
@johnkslang I'm not sure if you need to see + merge this (I could merge it myself) but assigning to you because I don't want to tread on any toes!

PR #53, opened 2019-02-07
Remove support for gl_MeshViewCountNV/gl_MeshViewIndicesNV from task shaders

sparmarNV:
This was a spec bug which incorrectly added support for these builtins in task shaders.
Note that, our glslang implementation already support them in mesh shaders alone.

CC @nvpbrown

### DIFF
--- a/extensions/nv/GLSL_NV_mesh_shader.txt
+++ b/extensions/nv/GLSL_NV_mesh_shader.txt
@@ -23,8 +23,8 @@ Status
 
 Version
 
-    Last Modified Date:     October 22, 2018
-    NVIDIA Revision:        6
+    Last Modified Date:     February 7, 2019
+    NVIDIA Revision:        7
 
 Dependencies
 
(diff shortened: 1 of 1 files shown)


nvpbrown:
I had discussed this with @sparmarNV by email before he submitted this, but now realize that I misunderstood what he was asking about (below).

I'm now not convinced that we should proceed with this change, though it is apparently consistent with our glslang implementation.

--

I didn’t look at the email carefully enough.  For some reason, I thought you were talking about per-view *outputs*.  Now that I reviewed your spec edits, I realized that these are inputs.  I don’t think there’s any reason why the task shader can’t use them.  There might be use cases where this is a completely reasonable thing to do – e.g., have the task shader pre-compute some properties for each view and store in the task memory for the mesh shaders to read.

sparmarNV:
Based on Pat's comment, this change is not needed.
I will update the glslang implementation to allow these builtins in task shaders.
Closing this PR.

PR #54, opened 2019-02-18
Clarify how buffer reference types interact with const and constant expressions, and which operators are supported

jeffbolznv:
(set WIP just to give time for review)


johnkslang:
Looks good to me.

I wondered a little if there is a distinction between a reference and what object the reference is too, but it seems not. (E.g., the reference itself is not transparent, but what it's a reference too is transparent.)

When you de-WIP, let me know.

jeffbolznv:
@johnkslang I've removed WIP.

PR #61, opened 2019-03-15
Add GL_EXT_buffer_reference2 extension

jeffbolznv:
This is a GLSL-only followon to GL_EXT_buffer_reference that makes it easier to work with arrays of structures in memory, originally requested by @pixeljetstream.

I have a glslang implementation pending, I will push it soon.


jeffbolznv:
Removed WIP, I'd like to merge this soon.

PR #74, opened 2019-06-13
Add gl_SemanticsVolatile to GL_KHR_memory_scope_semantics

jeffbolznv:


extensions/khr/GL_KHR_memory_scope_semantics.txt
```diff
@@ -25,7 +25,7 @@ Status
 
 Version
 
-    Last Modified Date: 30-Aug-2018
+    Last Modified Date: 13-Jun-2019
     Revision: 1
```
johnkslang: Need to bump to Revision 2.

jeffbolznv: done

Issue #77, opened 2019-07-14
Fragment shader invocation generation for uncovered neighbors in subgroup quads


pow2clk:
Hello again Khronos!

I was intrigued by this section of extensions/khr/GL_KHR_shader_subgroup.txt (lines 606-611):

        If a primitive covers a fragment at (x, y), its fragment shader
        invocation will be in a quad with fragment shader invocations
        corresponding to the three neighboring pixels at (x + 1, y), (x, y + 1),
        and (x + 1, y + 1).  These four invocations are arranged in a 2x2 grid,
        that make up the quad.  If the neighbors of a fragment are not covered
        by the primitive, fragment shader invocations will still be generated.

What isn't clear to me from the existing text is the nature of the fragment shader invocations generated for uncovered neighbors. More specifically, I'd like to know how to distinguish quads or, barring that, fragment shader invocations that include such uncovered neighbors from those that don't.

A "knowledgeable source" (okay fine it was @nsubtil) suggested it might be done with `subgroupBallot(true)` or `gl_HelperInvocation`, but as the text neither says the invocations are inactive nor that they are helpers, it is not at all clear.

pow2clk:
Trying to decide whom to tag in this. It looks like Neil Henning has moved away from khronos stuff if his webpage is to believed. I guess I'll just go with @johnkslang and @oddhack and hope I'm not annoying them.

pow2clk:
For what it's worth, my friends at the DXC project have some documentation that says that the additional "lanes" are "helpers" assuming helper there means the same thing as helper here (It does right? Please tell me it does!) that would seem to resolve the question assuming GLSL wants to conform to the dark side. I would still strongly advocate more specific language in the spec to clarify that.

see https://github.com/Microsoft/DirectXShaderCompiler/wiki/Wave-Intrinsics

Specifically: 

    Helper Lane: A lane which is executed solely for the purpose of gradients in pixel shader quads. 
    The output of such a lane will be discarded, and so not rendered to the destination surface.

and

    Since these routines rely on quad-level values, they assume that all lanes in the quad are made
    active, even helper lanes (those that are masked from final writes).

pow2clk:
Resolved via #78. Thanks all!

PR #78, opened 2019-07-14
Clarify nature of added fragments to quads

pow2clk:
When a primitive only partially covers a quad of fragments, additional
fragment shader invocations are generated to fill it out. According to
other documentation, these are helper invocations. This clarifies that
they are helpers, which suggests how the case where the quad is not
completely composed of covered fragments can be detected.

Resolves issue #77

### DIFF
--- a/extensions/khr/GL_KHR_shader_subgroup.txt
+++ b/extensions/khr/GL_KHR_shader_subgroup.txt
@@ -50,8 +50,8 @@ Status
 
 Version
 
-    Last Modified Date: 17-Dec-2018
-    Revision: 7
+    Last Modified Date: 14-Jul-2019
+    Revision: 8
 
 Number
 
@@ -450,7 +450,8 @@ Additions to Chapter 3 of the OpenGL Shading Language Specification
           if the local workgroup size is not a multiple of the subgroup size.
 
         In fragment shaders, helper invocations participate in subgroup
-        operations.
+        operations. Such as when additional fragment shader invocations are
+        created to fill out a quad partially covered by a primitive.
 
         For each active invocation within a subgroup that reaches the same
         dynamic instance of a subgroup built-in function, all active invocations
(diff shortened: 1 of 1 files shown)


extensions/khr/GL_KHR_shader_subgroup.txt
```diff
@@ -450,7 +450,8 @@ Additions to Chapter 3 of the OpenGL Shading Language Specification
           if the local workgroup size is not a multiple of the subgroup size.
 
         In fragment shaders, helper invocations participate in subgroup
-        operations.
+        operations. Such as when additional fragment shader invocations are
+        created to fill out a quad partially covered by a primitive.
```
jeffbolznv: "Such as" makes this sound like a sentence fragment. Is this edit really necessary, or is the other one enough by itself?

pow2clk: It is a sentence fragment. That's bad. 

It's not strictly necessary, but the original sentence seemed a bit sparse on its own given that it didn't say what circumstances would necessitate helper invocation participation.

My preference would be to say: `helper invocations participate in subgroup operations when . . . ` but I'm not sure I know that complete list of "whens", so I thought I'd add the one I was confident of. Perhaps the whens take too long to explain?

jeffbolznv: We've had some internal discussions about questions like "when are helper invocations guaranteed to exist", "can they dynamically come in and out of existence", "is there a way to force them to exist at a given point in a program", things like that, and they don't have easy answers. We haven't gotten far enough to be able to state anything definitive about it, so I think we should continue to err on the side of being vague.

pow2clk: Fair enough. Unfortunate that clarity is not possible, but better to be vague than misleading.

PR #89, opened 2019-09-19
Update GLSL_EXT_demote_to_helper_invocation dependencies

cmarcelo:
Change dependency with "GL_KHR_vulkan_glsl" from requirement to
interaction.  Expand the GLSL spec dependencies.

The goal is make clear that this extension can also be used in OpenGL.

### DIFF
--- a/extensions/ext/GLSL_EXT_demote_to_helper_invocation.txt
+++ b/extensions/ext/GLSL_EXT_demote_to_helper_invocation.txt
@@ -18,8 +18,8 @@ Status
 
 Version
 
-    Last Modified Date:         June 06, 2019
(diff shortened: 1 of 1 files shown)


extensions/ext/GLSL_EXT_demote_to_helper_invocation.txt
```diff
@@ -27,7 +27,16 @@ Number
```
dgkoch: You should update the revision and last modified date above, and the revision history at the bottom of the file.

cmarcelo: Done.

PR #97, opened 2019-10-16
Typo "multiplies" to muliplications

abhilash1910:


pdaniell-nv:
I think "multiplies" is the better term. From https://en.wiktionary.org/wiki/multiply:
```
multiply (plural multiplies)
    (computer science) An act or instance of multiplying.
```

PR #110, opened 2020-03-10
EXT_debug_printf - make escape sequences better match C/C++

jeffbolznv:


johnkslang:
This is the correct correction to match C++ rules.

PR #112, opened 2020-03-17
Forbid memoryBarrierAtomicCounter() in GL_KHR_vulkan_glsl

jeffbolznv:


johnkslang:
I think we either need this just in GLSL 4.6.7, or both.  Often, after something is incorporated, and the core spec evolves, it supersedes the original extension spec.

PR #120, opened 2020-05-06
GL_EXT_vulkan_glsl_relaxed, for review and feedback

mbechard:


extensions/ext/GL_EXT_vulkan_glsl_relaxed.txt
```diff
@@ -0,0 +1,177 @@
...
+    
+    This extension can be applied to OpenGL GLSL versions 1.40
+    (#version 140) and higher.
+
+    This extension can be applied to OpenGL ES ESSL versions 3.10
+    (#version 310) and higher.
+
+    All these versions map GLSL/ESSL semantics to the same SPIR-V 1.0
+    semantics (approximating the most recent versions of GLSL/ESSL).
+
```
johnkslang, line 43: There should also be a paragraph about what versions of what specifications this is written against.  This is usually a quite recent specification version, and independent of the statements of how back the extension can be applied.

extensions/ext/GL_EXT_vulkan_glsl_relaxed.txt
```diff
@@ -0,0 +1,177 @@
...
+    'gl_AtomicBufferBlock'. The block name will suffixed with '_<bindingNumber>'.
+    Where <bindingNumber> is the original binding specified for the
+    atomic_uints contained within that block. The actual binding of the
+    block may be different than what was originally specified in the shader.
+    Reflection should be used to find the final binding locations of these
+    blocks.
+   
+Changes to Chapter 3 of the OpenGL Shading Language Specification
+
+    Move the following keywords in section 3.6 Keywords back out of
```
johnkslang, line 126: Here is where it is important to know what specification these modification instructions are against.

extensions/ext/GL_EXT_vulkan_glsl_relaxed.txt
mbechard: Just so I'm clear. Are you saying specify which version of GL_KHR_vulkan_glsl the spec is against, or also the GLSL version?

johnkslang: It depends what specification you are writing it against.

Given that this has been incorporated into the GLSL specification, I think it's best to write against that:  The OpenGL Shading Language, Version 4.60.7.

Then, each normative statement is about what to change in that spec.

mbechard: Since this extension relies heavily on the changes in GL_KHR_vulkan_glsl, is there an example (another extension somewhere that does similar) of how to reconcile both the changes that this extension wants from GL_KHR_vulkan_glsl as well as the changes it reverses.
Should I basically copy/paste all the change-entries from GL_KHR_vulkan_glsl and then edit them to be what they would be when both this extension and GL_KHR_vulkan_glsl are active? That way this extension includes both edits, or is there a different way of going about formatting this?

johnkslang: I recommend writing relative to "The OpenGL Shading Language, Version 4.60.7."

You would just say what edits should be made there, and independently state "where the modified features also exist in previous versions, these changes are also supported in previous versions".

johnkslang: GL_EXT_vulkan_glsl itself is an example of how to do that.

mbechard: GL_EXT_vulkan_glsl does not contain any paragraph stating what versions of what specification it's written against though, no? Can you be specific about a section from GL_EXT_vulkan_glsl whose formatting/writing style this extension is failing to follow?

Sorry, I'm really struggling to understand what you are asking me to do here. Other than adding a line stating "This was written against The OpenGL Shading Language, Version 4.60.7", I'm unclear what you would like me to edit, and how it should be edited.

I'll add that paragraph, but if there is more I'm missing maybe can you could do a single example edit to this spec to show what you are looking for? Thanks and sorry for not understanding. Thanks

mbechard: Took another stab at how the edits are formatted, and went through the 4.60.7 spec to find other spots that need to be edited by this spec. Hoping this is more in-line with what you are thinking.

johnkslang:
This seems a decent specification to compare an implementation against.

Any other input from anyone?

johnkslang:
You should also include a file change to the README to get this listed in the index on the top page.

mbechard:
> You should also include a file change to the README to get this listed in the index on the top page.

Done, thanks

mbechard:
I do wonder if something needs to be done with gl_FragColor here too. The GLSL spec currently states that gl_FragColor is only available in compatibility profile, and that compatibility profile isn't available when generating SPIR-V. So if I'm correct then the only way to use gl_FragColor with SPIR-V is when doing ESSL->SPIR-V, correct?
In my experience there are lots of shaders out there that write to gl_FragColor via the compatibility profile in GL. So maybe allowing gl_FragColor to be an implicit write to location=0 could be useful as well in this spec.

johnkslang:
`gl_FragColor` has some built-in behavior for broadcast across multiple targets.  Maybe it is better to get away from using it, rather than have it only partially emulated?

johnkslang:
Khronos would like to keep this PR open for 2 more weeks to solicit more feedback.

mbechard:
> `gl_FragColor` has some built-in behavior for broadcast across multiple targets. Maybe it is better to get away from using it, rather than have it only partially emulated?

GL_EXT_vulkan_glsl already does away with the broadcast behavior though, and I think it's good to keep that change. In my experience a lot of my customers shaders end up using gl_FragColor, often due to coming from ESSL sources. So the new change would just be to allow gl_FragColor to work as location=0 for GLSL shaders as well as ESSL shaders.

mbechard:
I'm not actually sure if there is a way to use gl_FragColor in combination with SPIR-V with ESSL shaders right now anyways, due to ESSL version requirements. Is that correct? If that's the case then let's drop the gl_FragColor idea.

mbechard:
Hey, just checking in again on if anything else needs to be done for this?

PR #124, opened 2020-05-22
Update spec with new layout qualifier.

alelenv:
Spec update for primitive_culling.
@dgkoch for initial review

### DIFF
--- a/extensions/ext/GLSL_EXT_ray_flags_primitive_culling.txt
+++ b/extensions/ext/GLSL_EXT_ray_flags_primitive_culling.txt
@@ -100,12 +100,12 @@ Additions to Chapter 4 of the OpenGL Shading Language Specification
 
     4.4.10 Primitive Culling Layout Qualifier
 
-    This layout qualifier can only be applied only at global scope
+    This layout qualifier can only be applied at global scope
     and without an interface-qualifier.
 
     It is a compile-time error to use this qualifier with an
-    interface-qualifier, on a block or member, or with any other
-    layout qualifiers.
+    interface-qualifier, on an interface block or member of an interface
+    block, or with any other layout qualifiers.
 
 
 Additions to Chapter 7 of the OpenGL Shading Language Specification
--- a/extensions/ext/GLSL_EXT_ray_flags_primitive_culling.txt
+++ b/extensions/ext/GLSL_EXT_ray_flags_primitive_culling.txt
@@ -21,8 +21,8 @@ Status
 
 Version
 
-    Last Modified Date: 2020-04-01
-    Revision: 1
+    Last Modified Date: 2020-05-21
+    Revision: 2
 
 Dependencies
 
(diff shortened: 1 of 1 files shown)


extensions/ext/GLSL_EXT_ray_flags_primitive_culling.txt
```diff
@@ -73,6 +74,40 @@ Modifications to the OpenGL Shading Language Specification, Version 4.60
       #define GL_EXT_ray_flags_primtive_culling          1
 
 
+Additions to Chapter 4 of the OpenGL Shading Language Specification
+(Variables and Types)
+
+    Modify Section 4.4, Layout Qualifiers
```
dgkoch, line 80: Should probably add something to the SPIR-V mapping section?

I think the primitive_culling layout maps to the RayTraversalPrimitiveCullingProvisionalKHR capability.

extensions/ext/GLSL_EXT_ray_flags_primitive_culling.txt
```diff
@@ -73,6 +74,40 @@ Modifications to the OpenGL Shading Language Specification, Version 4.60
...
+
+    4.4.10 Primitive Culling Layout Qualifier
+
+    This layout qualifier can only be applied at global scope
+    and without an interface-qualifier.
+
+    It is a compile-time error to use this qualifier with an
+    interface-qualifier, on an interface block or member of an interface
+    block, or with any other layout qualifiers.
+
```
dgkoch, line 109: We should probably add something stating what this layout is actually *for*. 

Ie - to indicate that triangle or AABB primitive culling is being used and must be specified when using the SkipTriangles or SkipAABBs ray flags or ray flag constants?

extensions/ext/GLSL_EXT_ray_flags_primitive_culling.txt
```diff
@@ -73,6 +74,40 @@ Modifications to the OpenGL Shading Language Specification, Version 4.60
...
+
```
dgkoch, line 110: should it be a compile time error if either of the new constants are used without having this declared?

extensions/ext/GLSL_EXT_ray_flags_primitive_culling.txt
alelenv: Updated.

extensions/ext/GLSL_EXT_ray_flags_primitive_culling.txt
alelenv: Updated

extensions/ext/GLSL_EXT_ray_flags_primitive_culling.txt
alelenv: Cannot really prevent someone from using values '256U' anyway, so erroring out for one case doesnt make sense . All this qualifier does is enables a capability which defines valid behavior for the flag values? Otherwise results are undefined

dgkoch: fair enough.

dgkoch:
@johnkslang did you want to review this?

johnkslang:
Looks good.

PR #150, opened 2021-01-25
Add GL_EXT_shared_memory_block

cmarcelo:


helixd-2k18:
About `shared / packed / std140 / std430` line 185... 
Will there be "scalar" support for shared blocks?

cmarcelo:
> About `shared / packed / std140 / std430` line 185...
> Will there be "scalar" support for shared blocks?

Yes, it is already there. If both this and GL_EXT_scalar_block_layout is used, and if the target SPIR-V environment supports it.

The only one we have at the moment is Vulkan, which has a feature bit to describe whether scalar layout is supported or not.

seanbaxter:
This extension can't be used when specialization constants size shared memory capacity, as OpMemberDecorate can only source literals, not Constant expressions, and Workgroup variables are now storage classes requiring layout.

PR #154, opened 2021-03-04
Update GLSL_EXT_ray_query.txt

TylerNowicki:
Clarify rayQueryGetIntersectionCandidateAABBOpaqueEXT description

### DIFF
--- a/extensions/ext/GLSL_EXT_ray_query.txt
+++ b/extensions/ext/GLSL_EXT_ray_query.txt
@@ -432,7 +432,7 @@ Additions to Chapter 8 of the OpenGL Shading Language Specification
 
         bool rayQueryGetIntersectionFrontFaceEXT(rayQueryEXT q, bool committed);
 
-    Returns whether current intersection is front face triangle.
+    Returns whether current intersection is a front facing triangle.
 
     If <committed> is 'true'  returns value for committed intersection.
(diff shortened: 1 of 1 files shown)


extensions/ext/GLSL_EXT_ray_query.txt
```diff
@@ -442,8 +442,8 @@ Additions to Chapter 8 of the OpenGL Shading Language Specification
 
         bool rayQueryGetIntersectionCandidateAABBOpaqueEXT(rayQueryEXT q);
 
-    Returns whether current candidate intersection is opaque geometry
-    or not.
+    Returns whether current candidate intersection is opaque AABB or
```
gnl21, line 445:
```suggestion
    Returns whether current candidate intersection is an opaque AABB or
```

fluppeteer:
I know this is pedantic, but I do have to look twice at anything that says "returns whether something is true or not" to confirm it's not "returns whether A is true or B is true" for independent A and B. I probably hang around too many people who use boolean logic answers to either/or questions.

Could we remove the "or not"s, or say "returns true if the current candidate intersection is opaque"?

dgkoch:
We should also update the changelog at the bottom and version at the top of the file.

dgkoch (approved):
LGTM - Thanks!

PR #163, opened 2021-06-19
add GLSL_HUAWEI_subpass_shading

WyvernWang:


extensions/huawei/GLSL_HUAWEI_subpass_shading.txt
```diff
@@ -0,0 +1,86 @@
...
+
+    Subpass-input types are only available in fragment and subpass shading shaders.
+
+Changes to Chapter 8 of The OpenGL Shading Language Specification, Version 4.60
+
+    Modify Section 8.18, (Subpass-Input Functions)
+
+    Change the "Subpass-input functions are only available when targeting a Vulkan fragment stage." to 
+
+    Subpass-input functions are only available when targeting a Vulkan fragment and subpass shading stage.
```
gnl21, line 67: I think this one should be "fragment *or* subpass shading stage"

extensions/huawei/GLSL_HUAWEI_subpass_shading.txt
```diff
@@ -0,0 +1,86 @@
...
+
+    This extension is written against revision 7 of the OpenGL Shading Language
+    version 4.60, dated July 10, 2019.
+
+    This extension interacts with revision 43 of the GL_KHR_vulkan_glsl
+    extension, dated October 25, 2017.
+
+Overview
+
+    This extension allows a subpass shading shader to delcare subpass inputs and use subpass input functions.
```
gnl21, line 39: I find "subpass shading shader" quite awkward. Do you prefer that term to the simpler "subpass shader"?

extensions/huawei/GLSL_HUAWEI_subpass_shading.txt
```diff
@@ -0,0 +1,86 @@
...
+
+    Last Modified Date: 2021-04-26
+    Revision: 1
+
+Dependencies
+
+    This extension can be applied to OpenGL GLSL versions 4.60
+    (#version 460) and higher.
+
+    This extension is written against revision 7 of the OpenGL Shading Language
```
gnl21, line 31: You don't give any interactions with the ES version of the spec. Is it your intention that the ES version of the shading language should be compatible with these shaders?

gnl21 (commented):
It would be nice to add a section at the end of section 2 of the spec describing what subpass shading is (at the moment, each shader stage has a short description there).

extensions/huawei/GLSL_HUAWEI_subpass_shading.txt
WyvernWang: Thanks, I fixed in https://github.com/KhronosGroup/GLSL/pull/163/commits/1114f6bed5d92f19481adecdad26a081fa1b4e09

extensions/huawei/GLSL_HUAWEI_subpass_shading.txt
WyvernWang: I've tried but finally I still have to accept it since it pulls the strings to other naming like pipelines, stages.

extensions/huawei/GLSL_HUAWEI_subpass_shading.txt
WyvernWang: Okay, I added them in https://github.com/KhronosGroup/GLSL/pull/163/commits/1114f6bed5d92f19481adecdad26a081fa1b4e09

extensions/huawei/GLSL_HUAWEI_subpass_shading.txt
gnl21: OK, fair enough.

WyvernWang:
> It would be nice to add a section at the end of section 2 of the spec describing what subpass shading is (at the moment, each shader stage has a short description there).

Thanks for the tip, I've done this.

extensions/huawei/GLSL_HUAWEI_subpass_shading.txt
```diff
@@ -0,0 +1,147 @@
...
+    This extension can be applied to OpenGL GLSL versions 4.60
+    (#version 460) and higher.
+
+    This extension can be applied to OpenGL ES ESSL versions 3.20
+    (#version 320) and higher.
+    
+    This extension is written against revision 7 of the OpenGL Shading Language
+    version 4.60, dated July 10, 2019.
+
+    This extension is written against revision 6 of the OpenGL ES ESSL
```
gnl21, line 37: I don't think you need to include the actual changes twice, you can just say that the extension can be applied to ES, but that it's written against the desktop version. If there are any significant differences between desktop and ES then you can note them at the end in a section of "Interactions with Open GL ES". Sorry to have misled you on this.

extensions/huawei/GLSL_HUAWEI_subpass_shading.txt
```diff
@@ -0,0 +1,147 @@
...
+    
+    The subpass shading processor is a programmable unit that operates the subpass inputs. It runs independently 
+    from the other shader processors. Compilation units written in the OpenGL Shading Language to run on this
+    processor are called subpass shading shaders. When a set of subpass shading shaders are successfully compiled 
+    and linked, they result in a subpass shading shader executable that runs on the subpass shading processor.
+
+    A subpass shading shader is almost identical to a compute shader except the following
+    differences.
+        
+        * A subpass shading shader must be dispatched in a dedicate subpass
```
gnl21, line 71:
```suggestion
        * A subpass shading shader must be dispatched in a dedicated subpass
```

extensions/huawei/GLSL_HUAWEI_subpass_shading.txt
```diff
@@ -0,0 +1,147 @@
...
+        * A subpass shading shader is allowed to read fragment from input attachment
```
gnl21, line 72:
```suggestion
        * A subpass shading shader is allowed to read fragments from input attachments
```

extensions/huawei/GLSL_HUAWEI_subpass_shading.txt
```diff
@@ -0,0 +1,147 @@
...
+
+Changes to Chapter 12 of The OpenGL Shading Language Specification, Version 4.60
+
+    Modify Section 12.2.4, (Vulkan Only: Samplers, Images, Textures, and Buffers)
+
+    Change the "Subpass inputs are read through a new set of types, available only to fragment shaders:" to 
+
+    Subpass inputs are read through a new set of types, available only to fragment or subpass shading shaders:
+
+Additions to Chapter 2 of the OpenGL ES ESSL versions 3.20
```
gnl21, line 99: As above, I would remove these sections about ES and only note where there are differences from desktop, if any.

oddhack:
Is it likely this will get merged soon? It's referred from the VK_HUAWEI_subpass_shading Vulkan spec appendix and the link is generating 404 errors which affect webcrawlers, so if it's likely to be delayed for a while, we should temporarily remove the dependency from Vulkan.

gnl21:
There are only editorial questions open, so I'm happy to merge it whenever @WyvernWang wants to. I've left some suggestions for improvements, but I don't think it should take long.

WyvernWang:
@gnl21 I've done the modification based on your suggestion. If ok, please go ahead the merge process.

gnl21:
@WyvernWang, I think that it would be better not to write out all the changes twice, once for ES and once for desktop. I think the way this is normally done is that the introduction says which versions are required for each of desktop and ES, then it says one version that the extension is written against (for example desktop 460). You include only the changes to that version and then note if there are any differences when applied to ES.

I don't think this is critical, so if you confirm that you want it merged as-is then I'll merge it, but I think it would be nicer to resolve that.

WyvernWang:
> @WyvernWang, I think that it would be better not to write out all the changes twice, once for ES and once for desktop. I think the way this is normally done is that the introduction says which versions are required for each of desktop and ES, then it says one version that the extension is written against (for example desktop 460). You include only the changes to that version and then note if there are any differences when applied to ES.
> 
> I don't think this is critical, so if you confirm that you want it merged as-is then I'll merge it, but I think it would be nicer to resolve that.

Well.. it looks like I try to edit and make a new branch fail to merge the fix...
Let me look what happened...

extensions/huawei/GLSL_HUAWEI_subpass_shading.txt
```diff
@@ -0,0 +1,105 @@
...
+
+    Modify Section 4.1.7, (Opaque Types)
+
+    Change the "Subpass-input types are only available in fragment shaders." to 
+
+    Subpass-input types are only available in fragment or subpass shading shaders.
+
+Changes to Chapter 8 of The OpenGL Shading Language Specification, Version 4.60 
+
+    Modify Section 8.18, (Subpass-Input Functions),
```
gnl21, line 83:
```suggestion
    Modify Section 8.18, (Subpass-Input Functions)
```

gnl21 (approved):
Looks good now, thanks.

One stray comma has appeared in this last change, which it would be good to remove, but not enough to hold things up for. I'll merge this when that's fixed, or I'll merge it tomorrow anyway.

PR #186, opened 2022-06-01
Add MSAA interaction note to GLSL_EXT_fragment_shader_barycentric

stu-s:


jekstrand:
We should probably also say something about whether or not you can use qualifiers.  From today's meeting, I think that's still a tad unclear so we may need to wait until we've confirmed what different implementations support.

stu-s:
Yeah, I think that we should wait a bit with the qualifier stuff.  I wasn't sure whether it was worth pushing this bit now, since it's "better than nothing", and then tweaking it once we've sorted out qualifiers?

jekstrand:
Yeah, that's fine with me I guess, as long as we don't close the internal issue until both pieces are resolved.

jekstrand:
Looks good to me.

stu-s:
@johnkslang I think this PR is good to go now.

stu-s:
@johnkslang @gnl21 Is anyone able to sign this off and merge it, please?

PR #195, opened 2022-10-04
Add GLSL_EXT_integer_dot_product

kpet:
Signed-off-by: Kevin Petit <kevin.petit@arm.com>
Change-Id: Ifc55b44217819cf5c6998eeb4aedcc8d7597e65d

### DIFF
--- a/extensions/ext/GLSL_EXT_integer_dot_product.txt
+++ b/extensions/ext/GLSL_EXT_integer_dot_product.txt
@@ -1,0 +1,288 @@
+Name
+
+    EXT_integer_dot_product
+
+Name Strings
+
+    GL_EXT_integer_dot_product
+
+Contact
+
+    Jeff Bolz, NVIDIA (jbolz 'at' nvidia.com)
+
+Contributors
+
+    Jeff Bolz, NVIDIA (jbolz 'at' nvidia.com)
+    Kevin Petit, Arm Ltd. (kevin.petit 'at' arm.com)
+    Tobias Hector, AMD
+    Graeme Leese, Broadcom Corporation
+    Georg Lehmann
+
+Status
+
+    Complete.
+
+Version
+
+    Last Modified:      June 7, 2023
+    Revision:           1
+
+Dependencies
+
+    This extension can be applied to OpenGL GLSL versions 4.50
+    (#version 450) and higher.
+
+    This extension can be applied to OpenGL ES ESSL versions 3.00
+    (#version 300) and higher.
+
+    This extension is written against the OpenGL Shading Language
+    Specification, version 4.60, dated July 23, 2017.
+
+    This extension interacts with
+    GL_EXT_shader_explicit_arithmetic_types_int8,
+    GL_EXT_shader_explicit_arithmetic_types_int16 and
+    GL_EXT_shader_explicit_arithmetic_types_int64.
+
+Overview
+
+    This extension adds mixed-precision integer dot product instructions.
+
+    Mapping to SPIR-V
+    -----------------
+
+    For informational purposes (non-normative), the following is an
+    expected way for an implementation to map GLSL constructs to SPIR-V
+    constructs, the SPIR-V capabilities they require, and the Vulkan
+    feature bits those capabilities require.
+
+    All functions introduced in this extension require support for
+    the shaderIntegerDotProduct Vulkan feature and DotProductKHR SPIR-V
+    capability.
+
+    The dotPacked4x8EXT and dotPacked4x8AccSatEXT functions require support for
+    the DotProductInput4x8BitPackedKHR SPIR-V capability.
+
+    All functions with i8vec4 or u8vec4 input vectors require support for the
+    DotProductInput4x8BitKHR SPIR-V capability.
+
+    All functions with input vectors of other types require support for the
+    DotProductInputAllKHR SPIR-V capability.
+
+    Function described in this extension map to SPIR-V instructions as follows:
+
+      dot(Packed4x8)EXT(unsigned, unsigned)                 -> Opcode: OpUDotKHR
+      dot(Packed4x8)EXT(signed, unsigned)                   -> Opcode: OpSUDotKHR
+      dot(Packed4x8)EXT(unsigned, signed)                   -> Opcode: OpSUDotKHR
+      dot(Packed4x8)EXT(signed, signed)                     -> Opcode: OpSDotKHR
+      dot(Packed4x8)AccSatEXT(unsigned, unsigned, unsigned) -> Opcode: OpUDotAccSatKHR
+      dot(Packed4x8)AccSatEXT(signed, unsigned, signed)     -> Opcode: OpSUDotAccSatKHR
+      dot(Packed4x8)AccSatEXT(unsigned, signed, signed)     -> Opcode: OpSUDotAccSatKHR
+      dot(Packed4x8)AccSatEXT(signed, signed, signed)       -> Opcode: OpSDotAccSatKHR
+
+Modifications to the OpenGL Shading Language Specification, Version 4.60
+
+    Including the following lines in a shader can be used to control the
+    language features described in this extension:
+
+      #extension GL_EXT_integer_dot_product : <behavior>
+
+    where <behavior> is as specified in section 3.3.
+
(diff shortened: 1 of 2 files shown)


extensions/ext/GLSL_EXT_integer_dot_product.txt
```diff
@@ -0,0 +1,287 @@
...
+
+      uint64_t dotAccSatEXT(u64vec3 a, u64vec3 b, uint c)
+      int64_t dotAccSatEXT(i64vec3 a, u64vec3 b, int c)
+      int64_t dotAccSatEXT(u64vec3 a, i64vec3 b, int c)
+      int64_t dotAccSatEXT(i64vec3 a, i64vec3 b, int c)
+
+      uint64_t dotAccSatEXT(u64vec4 a, u64vec4 b, uint c)
+      int64_t dotAccSatEXT(i64vec4 a, u64vec4 b, int c)
+      int64_t dotAccSatEXT(u64vec4 a, i64vec4 b, int c)
+      int64_t dotAccSatEXT(i64vec4 a, i64vec4 b, int c)
```
DadSchoorse, line 249: The accumulator `c` should be 64bit too for these.

kpet: Good catch, thanks! Now fixed. (I've added you to the list of contributors but couldn't find your affiliation. Feel free to comment with a suggestion if you want one added.)

nihui:
I wonder if there is currently good and handy type conversion function for packed4x8 type in integer_dot_product, like packHalf2x16() unpackHalf2x16() for packed fp16.
Shouldn't there be packInt4x8EXT() packUInt4x8EXT() unpackInt4x8EXT() unpackUInt4x8EXT() for packed 8bit integer ?

Existing conversions such as packSnorm4x8() will lose precision due to norm and are not suitable for use in computational tasks.

extensions/ext/GLSL_EXT_integer_dot_product.txt
```diff
@@ -0,0 +1,288 @@
...
+
+    Last Modified:      October 3, 2022
+    Revision:           1
+
+Dependencies
+
+    This extension can be applied to OpenGL GLSL versions 4.50
+    (#version 450) and higher.
+
+    This extension can be applied to OpenGL ES ESSL versions 3.20
```
gnl21, line 35: Is there any reason why this requires ES3.2? Could it not be applied to version 300?

extensions/ext/GLSL_EXT_integer_dot_product.txt
```diff
@@ -0,0 +1,288 @@
...
+
+    All functions introduced in this extension require support for
+    the shaderIntegerDotProduct Vulkan feature and DotProductKHR SPIR-V
+    capability.
+
+    The dotPacked4x8EXT and dotPacked4x8AccSatEXT functions require support for
+    the DotProductInput4x8BitPackedKHR SPIR-V capability.
+
+    All functions with ivec8 or uvec8 input vectors require support for the
+    DotProductInput4x8BitKHR SPIR-V capability.
```
gnl21, line 66: The SPIR-V for this is specific to 4-component vectors, so there seems to be a mismatch. Should one or other be updated to match?

extensions/ext/GLSL_EXT_integer_dot_product.txt
```diff
@@ -0,0 +1,288 @@
...
+      int dotPacked4x8EXT(int a, uint b)
+      int dotPacked4x8EXT(uint a, int b)
+      int dotPacked4x8EXT(int a, int b)
+
+      If GL_EXT_shader_explicit_arithmetic_types_int8 is enabled:
+
+      uint dotEXT(u8vec2 a, u8vec2 b)
+      int dotEXT(i8vec2 a, u8vec2 b)
+      int dotEXT(u8vec2 a, i8vec2 b)
+      int dotEXT(i8vec2 a, i8vec2 b)
```
gnl21, line 124: It's a bit odd that 32 and 64 bit dot products are not extended but these lower-bit-depth variants extend to 32-bits first. SPIR-V has opcodes for 8 x 8 -> 8, etc. which may be faster than these 8x8 -> 32 variants. Do you think these are more useful to apps?

extensions/ext/GLSL_EXT_integer_dot_product.txt
```diff
@@ -0,0 +1,288 @@
...
+    (1) What overloads should this extension support?
+
+    RESOLVED: The VK_KHR_shader_integer_dot_product extension requires that all
+    data types accepted by the Vulkan implementation be usable with dot product
+    instructions. This extension provides overloads for all possible input
+    vector types. The return values are 32-bit integers, except for dot
+    products operating on vectors of 64-bit integers, as this is expected to be
+    the common case and it is easy for compilers to fold any subsequent
+    truncations or zero/sign-extensions into the dot product instruction as an
+    optimization.
```
gnl21, line 283: I see that this partially addresses my comment above, but it doesn't completely address it because the precision of the intermediate results is required to match the output precision, so a 32-bit operation and a truncate is not the same as a 16-bit operation.

gnl21:
> I wonder if there is currently good and handy type conversion function for packed4x8 type in integer_dot_product

Currently no, I don't think there is. One could be added but ...

> I admit that we could simulate this pack/unpack process with bit shift/or/and operations, but doing so is complex and limited and can be inefficient

I'm not sure why you say that they are limited, so perhaps you have something different in mind, but compilers should already understand that these are pack/unpack operations and be emitting good instruction sequences for them. I don't think we should expect any gain in efficiency from adding functions for this.

extensions/ext/GLSL_EXT_integer_dot_product.txt
kpet: I can't think of one, likely due to copy/paste. I'll change it to say 3.00.

extensions/ext/GLSL_EXT_integer_dot_product.txt
kpet: The GLSL spec should say `i8vec4` and `u8vec4`.

extensions/ext/GLSL_EXT_integer_dot_product.txt
kpet: I think it is because we require "wrap-around behaviour". Following the SPIR-V wording, the 32-bit operation would provide the correct result R.

gnl21: Yes, I think my original comment is just wrong, the operation + truncate does provide the same as the smaller operation, so this does resolve my comment above.

gnl21:
Thanks for the fixes. What's the state of implementations on this? Is it ready to be merged now?

kpet:
>Thanks for the fixes. What's the state of implementations on this? Is it ready to be merged now?

Thanks for the review! As far as I'm aware there are two prototypes from two different people, neither of which is complete. Neither author is able to continue work on their implementation. I expect this to be discussed in today's Vulkan ML teleconference and hopefully someone will be able to pick up the work. We could wait until we have a complete implementation to merge just in case implementation work turns up spec issues but I think this spec has been reviewed a good number of times by now so maybe publishing is helpful. I don't feel too strongly either way.

oddhack:
When resolving conflicts, please note that you will also need to move any changes in this branch to README.md, to corresponding asciidoc markup in README.adoc, and remove README.md .

jeffbolznv:
I've revived one of the in-progress implementations and (hopefully) finished it up: https://github.com/KhronosGroup/glslang/pull/3832

0cc4m:
@jeffbolznv Thank you for doing this, I was just about to look into the steps necessary for GLSL integer dot again when I found this.

jeremy-lunarg:
https://github.com/KhronosGroup/glslang/pull/3832 has been merged (perhaps prematurely?).  Where are we at with this PR? It'd be nice to land it since we landed the implementation in glslang.

jeffbolznv:
I think this just needs to be rebased and merged. I was going to raise it in the TSG meeting tomorrow.

kpet:
I've rebased the PR and updated `README.adoc`. This should be ready-to-go.

PR #199, opened 2023-02-18
Publish GLSL_QCOM_image_processing

jackohound:


jackohound:
@pdaniell-nv, is there something additional I need to do to get this onto the agenda and merged?

pdaniell-nv:
> @pdaniell-nv, is there something additional I need to do to get this onto the agenda and merged?

I think this extension is for Vulkan, so not something on my radar for the OpenGL/ES meetings. I think if @gnl21 is happy with it it can just be merged and published to the registry.

jackohound:
@gnl21 if you are happy with this vendor extension, can you merge it?

extensions/qcom/GLSL_QCOM_image_processing.txt
```diff
@@ -0,0 +1,231 @@
...
+    the same block from one texture is going to be used repeatedly across multiple
+    instructions, that texture should be specified as the referenceTexture.
+
+    * targetCoord and referenceCoord specify the texture coordinates (in texels)
+    of the bottom-left corner of each block.
+
+    * blockSize specifies the size (in texels) of the blocks to be compared.
+
+
+Modifications to the OpenGL ES Shading Language Specification, Version 3.10, dated 29 January 2016
```
gnl21, line 158: Are these the same additions for ES and desktop? It seems a shame to write the whole thing out twice. Perhaps just change this section to note any differences between the ES and desktop versions?

extensions/qcom/GLSL_QCOM_image_processing.txt
```diff
@@ -0,0 +1,231 @@
...
+    |   vec4 textureBlockMatchSSDQCOM(   | Block matching operation measures the      |
+    |       sampler2D target             | correlation (or similarity) of the target  |
+    |       uvec2     targetCoord,       | block and reference block.  TargetCoord    |
+    |       sampler2D reference,         | and refCoord specify the bottom-left corner|
+    |       uvec2     refCoord,          | of the block in target and reference       |
+    |       uvec2     blockSize)         | images. The error metric is the Sum of     |
+    |                                    | Square Differences (SSD).                  |
+    +------------------------------------+--------------------------------------------+
+
+    * <target> and <reference> specify the two input textures. If
```
gnl21, line 206: Here, for example, the text is different between ES and desktop, but it's not clear whether it's an intentional difference or just an editing problem.

extensions/qcom/GLSL_QCOM_image_processing.txt
```diff
@@ -0,0 +1,231 @@
...
+    instructions, that texture should be specified as the referenceTexture.
+
+    * For textureBlockMatchSADQCOM and textureBlockMatchSSDQCOM, <blockSize> specifies
+	  the size (in texels) of the blocks to be compared.
+
+
+Restrictions:
+
+    The <weight> texture used in the function textureWeightedQCOM cannot be used 
+    with any other texture functions.
```
gnl21, line 217: It would be nice to be more specific here. Are we talking about a compile-time error if the same sampler variable is used in multiple contexts, or undefined behaviour if it turns out that the same underlying texture is bound to multiple places?

extensions/qcom/GLSL_QCOM_image_processing.txt
```diff
@@ -0,0 +1,231 @@
...
+    |       sampler2D target             | correlation (or similarity) of the target  |
+    |       uvec2     targetCoord,       | block and reference block.  targetCoord    |
+    |       sampler2D reference,         | and refCoord specify the bottom-left corner|
+    |       uvec2     refCoord,          | of the block in target and reference       |
+    |       uvec2     blockSize)         | images. The error metric is the Sum of     |
+    |                                    | Absolute Differences (SAD).                |
+    +------------------------------------+--------------------------------------------+
+    |   vec4 textureBlockMatchSSDQCOM(   | Block matching operation measures the      |
+    |       sampler2D target             | correlation (or similarity) of the target  |
+    |       uvec2     targetCoord,       | block and reference block.  TargetCoord    |
```
gnl21, line 199: There's not much definition of what the return value might be here. Is there anything more you can say to give a clearer idea of how it should be used? Or a reference for where it could be found? This doesn't mention, for example, the ranges of values, whether higher or lower means more similar, what the different vector components mean.

jackohound: Updated with a little more information and a reference to the Vukan spec where the details are specified.

jackohound:
@gnl21, thamks for the suggested improvements.  I've incorporated your feedback and the extension is ready to publish.

PR #203, opened 2023-04-24
Add GL_EXT_shader_tile_image

janharaldfredriksen-arm:


extensions/ext/GLSL_EXT_shader_tile_image.txt
```diff
@@ -0,0 +1,306 @@
...
+    shader stage.
+
+    Modify Section 4.4, Layout Qualifiers (p. 63)
+
+    (add to the layout qualifier table, pg. 63-65)
+
+                                              Qualifier  Individual           Block     Allowed
+    Layout Qualifier                            Only      Variable   Block   Member   Interfaces
+    ----------------------------------------  ---------  ----------  -----  --------  -----------
+    non_coherent_color_attachment_readEXT        X          -         -        -     fragment in
```
janharaldfredriksen-arm, line 144: It has been suggested that the allowed interfaces here should instead be 'tileImageEXT'.

That would require some minor changes to the glslang implementation and (Khronos internal) CTS patches. I have not yet made this change, though could be persuaded to.

janharaldfredriksen-arm:
@gnl21 - can we merge this now?

PR #208, opened 2023-06-30
Initial revision of GL_KHR_cooperative_matrix

jeffbolznv:
I've imported this as-is from the private repo, so we can do final review here.

### DIFF
--- a/extensions/khr/GLSL_KHR_cooperative_matrix.txt
+++ b/extensions/khr/GLSL_KHR_cooperative_matrix.txt
@@ -1,0 +1,396 @@
+Name
+GL_KHR_cooperative_matrix
+
+Contact
+Neil Hickey, Arm (neil.hickey 'at' arm.com)
+
+Contributors
+Jeff Bolz, NVIDIA (jbolz 'at' nvidia.com)
+Kevin Petit, Arm (kevin.petit 'at' arm.com)
+
+Notice
+Copyright (c) 2022 The Khronos Group Inc. Copyright terms at http://www.khronos.org/registry/speccopyright.html
+
+Status
+Not yet released
+
+Version
+Last Modified: Feb 10, 2023
+Revision: 3
+
+
+
+Dependencies
+This extension can be applied to OpenGL GLSL versions 4.50
+(#version 450) and higher.
+This extension can be applied to OpenGL ES ESSL versions 3.20
+(#version 320) and higher.
+All these versions map GLSL/ESSL semantics to the same SPIR-V 1.5 semantics (approximating the most recent versions of GLSL/ESSL).
+
+This extension interacts with physical_storage_buffer,EXT_shader_explicit_arithmetic_types
+
+Overview
+
+This extension adds a new set of types known as "cooperative matrix" types,
+where the storage for and computations performed on the matrix are spread
+across a set of invocations such as a subgroup. These types give the
+implementation freedom in how to optimize matrix multiply operations.
+
+This extension introduces the types and built-in functions, but does not
+specify rules about what sizes/combinations are valid. This is left to
+the Vulkan extension specifications, and it is expected that different
+implementations may support different sizes. To help accommodate this,
+the dimensions of the cooperative types are parameterized and can be
+specialized via specialization constants.
+
+This extension introduces limited support for parameterized types, with
+the parameters specified as in C++ template syntax. The new built-in type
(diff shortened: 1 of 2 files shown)


extensions/khr/GLSL_KHR_cooperative_matrix.txt
```diff
@@ -0,0 +1,396 @@
...
+
+Contact
+Neil Hickey, Arm (neil.hickey 'at' arm.com)
+
+Contributors
+Jeff Bolz, NVIDIA (jbolz 'at' nvidia.com)
+Kevin Petit, Arm (kevin.petit 'at' arm.com)
+
+Notice
+Copyright (c) 2022 The Khronos Group Inc. Copyright terms at http://www.khronos.org/registry/speccopyright.html
```
jeffbolznv, line 12:
```suggestion
Copyright (c) 2022-2023 The Khronos Group Inc. Copyright terms at http://www.khronos.org/registry/speccopyright.html
```

extensions/khr/GLSL_KHR_cooperative_matrix.txt
```diff
@@ -0,0 +1,396 @@
+Name
+GL_KHR_cooperative_matrix
```
jeffbolznv, line 2: This file doesn't have the indentation that GLSL specs usually have. Should we fix?

extensions/khr/GLSL_KHR_cooperative_matrix.txt
```diff
@@ -0,0 +1,396 @@
...
+uvec2/uvec4.
+
+
+Revision History
+
+Revision 1
+
+- Internal revisions.
+
+Revision 2
```
jeffbolznv, line 392: I suggest to delete rev2-3 and only document revisions that occur after publication.

extensions/khr/GLSL_KHR_cooperative_matrix.txt
```diff
@@ -0,0 +1,396 @@
...
+of their element type and size. In the following functions, the generic type
+"coopmat" can accept an coopmat type with any type parameters. The
+"buf" arrays must be in either buffer storage or shared storage, and the
+array that is passed in can be sized or unsized.
+
+For all of these functions, for a given dynamic instance of the function
+call, all function parameters must be the same for all invocations in a
+given scope instance (where the scope is the scope the cooperative matrix
+type(s) were created with). All invocations in a given scope instance must
+be active or all must be inactive.
```
svenvh, line 188: This currently only applies to `coopMatLoad`, `coopMatStore` & `coopMatMulAdd`.  Do we need to say something similar in the "Expressions" section above?

extensions/khr/GLSL_KHR_cooperative_matrix.txt
```diff
@@ -0,0 +1,396 @@
...
+For informational purposes (non-normative), the following is an
+expected way for an implementation to map GLSL constructs to SPIR-V
+constructs:
+
+coopmat -> OpTypeCooperativeMatrixKHR
+coopmat constructor from scalar value -> OpConstantComposite
+coopmat constructor from coopmat -> Op*Convert, OpConvert*To*
+coopmat.length() -> OpCooperativeMatrixLength
+coopmat[i] -> OpCompositeExtract/OpCompositeInsert/OpAccessChain
++, -, *, / -> OpFAdd, OpFNegate/OpFSub, OpMatrixTimesScalar, OpFDiv
```
kpet, line 69: We might want to list **OpFMul** as well. Also, is it worth adding the mapping for integer operations?

kpet (changes requested):
A few minor things to address and this will IMO be good to go.

extensions/khr/GLSL_KHR_cooperative_matrix.txt
jeffbolznv: A similar question came up at https://github.com/KhronosGroup/SPIRV-Registry/issues/208, see my response there. I'm not planning to make a change in this spec unless we decide something different there.

jeffbolznv:
@kpet I think I've addressed all your feedback.

nihui:
May I ask when this PR will be merged, I look forward to testing this new extension in neural network inference in the downstream vkpeak and ncnn projects, relying on the implementation of glslang and this change :smiley:

kpet (approved):
@jeffbolznv LGTM, thanks!

PR #211, opened 2023-07-25
Rename GLSL_EXT_shader_tile_image.txt to GL_EXT_shader_tile_image.txt

PanGao-h:
Vulkan spec links to `GL_EXT_shader_tile_image` instead of `GLSL_EXT_shader_tile_image` and the name with GL prefix aligns with name strings.


jeffbolznv:
We've generally been trying to use GLSL for the file names to help indicate that they're not OpenGL extensions. I suggest fixing the link instead.

PanGao-h:
If that's the case, it's more reasonable to make the changes in spec instead. Close this and create a MR in doc repo then.

PR #218, opened 2023-10-03
Add procedure for handling opaque struct members

sbourasse:
GLSL specification states that opaque types "can only be declared as function parameters or in uniform-qualified variables." (GLSL 4.6 section 4.1.7) making them suitable structure members under OpenGL semantics. However, `GL_KHR_vulkan_glsl` explicitly forbids such behaviour.
Our proposal is to add support for such variables through `GL_EXT_vulkan_glsl_relaxed` so that existing OpenGL shader codebases can still be used without further changes.

We implemented this method within `glslang` already, so we will open a PR for it along this one, should it be approved.


mbechard:
This looks good to me

PR #221, opened 2023-11-12
GL_EXT_control_flow_attributes2

wooyoungqcom:
The GLSL extension "GL_EXT_control_flow_attributes" adds attributes to declare how loops should be lowered by a consumer (e.g., unroll and dont_unroll). But not all attributes available in SPIR-V 1.4 are made visible.

The PR is for an extension that adds the attributes missing in the GLSL extension GL_EXT_control_flow_attributes: they are min_iterations, max_iterations, iteration_multiple, peel_count, and partial_count.

Please note that the support for these attributes has already been available in glslangValidator.

### DIFF
--- a/extensions/ext/GL_EXT_control_flow_attributes2.txt
+++ b/extensions/ext/GL_EXT_control_flow_attributes2.txt
@@ -105,8 +105,8 @@ Add the following control-flow attributes to the table in Section 6.5 Control-Fl
 Add the following before "The *unroll*, *dont_unroll*, *flatten*, and *dont_flatten* attributes are"
 
     /I/ in *min_iterations(I)*/*max_iterations(I)*, /M/ in *iteration_multiple(M)*,
-    and /C/ in *peel_count(C)*/*partial_count(C)* must be a 32-bit positive signed
-    or unsigned non-specialization compile-time constant integral expression, or a
+    and /C/ in *peel_count(C)*/*partial_count(C)* must be a 32-bit signed
+    or unsigned non-specialization compile-time constant integral expression greater than 0, or a
     compile-time error results.
 
 Revision History
--- a/extensions/ext/GL_EXT_control_flow_attributes2.txt
+++ b/extensions/ext/GL_EXT_control_flow_attributes2.txt
@@ -25,7 +25,7 @@ Status
 
 Version
 
-    Last Modified Date: 31-Oct-2023
+    Last Modified Date: 30-Nov-2023
     Revision: 1
 
 Number
@@ -69,6 +69,7 @@ Overview
 
       [[unroll, min_iterations(2)]]  for (int i = 0; i < 8; ++i) { ... }
       [[unroll, min_iterations(2), max_iterations(4)]]  for (int i = 0; i < 8; ++i) { ... }
+      [[unroll, iteration_multiple(4)]]  for (int i = 0; i < 16; ++i) { ... }
       [[unroll, peel_count(3)]]  for (int i = 0; i < 11; ++i) { ... }
       [[unroll, partial_count(2)]]  for (int i = 0; i < 8; ++i) { ... }
 
@@ -119,4 +120,4 @@ Revision History
 
     Rev.    Date         Author          Changes
     ----  -----------    ------------    ----------------------------------------------
-    1     31-Oct-2023    Wooyoung Kim    initial draft
+    1     30-Nov-2023    Wooyoung Kim    initial draft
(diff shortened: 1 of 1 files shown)


extensions/ext/GL_EXT_control_flow_attributes2.txt
```diff
@@ -0,0 +1,123 @@
...
+      *iteration_multiple(M)* |    Loop     | the loop executes a multiple of a given number of iterations
+      *peel_count(C)*         |    Loop     | the loop be peeled by a given number of loop iterations
+      *partial_count(C)*      |    Loop     | the loop be partially unrolled by a given number of loop iterations
+
+
+Add the following before "The *unroll*, *dont_unroll*, *flatten*, and *dont_flatten* attributes are"
+
+    /I/ in *min_iterations(I)*/*max_iterations(I)*, /M/ in *iteration_multiple(M)*,
+    and /C/ in *peel_count(C)*/*partial_count(C)* must be a 32-bit positive signed
+    or unsigned non-specialization compile-time constant integral expression, or a
```
gnl21, line 109: I find this part slightly confusing. I assume from the following condition on M that this allows 0 for these values, but I would say that 0 is not a positive value. Could we clarify this, either by saying "non-negative" in place of "positive" or else by saying that it must be "greater than or equal to 0" instead?

extensions/ext/GL_EXT_control_flow_attributes2.txt
```diff
@@ -0,0 +1,123 @@
...
+
+Add the following control-flow attributes to the table in Section 6.5 Control-Flow Attributes
+
+       Attribute              | Relevant on | Intent
+      ------------------------+-------------+------------------------------------
+      *min_iterations(I)*     |    Loop     | the loop executes at least a given number of iterations
+      *max_iterations(I)*     |    Loop     | the loop executes at most a given number of iterations
+      *iteration_multiple(M)* |    Loop     | the loop executes a multiple of a given number of iterations
+      *peel_count(C)*         |    Loop     | the loop be peeled by a given number of loop iterations
+      *partial_count(C)*      |    Loop     | the loop be partially unrolled by a given number of loop iterations
```
gnl21, line 102: "the loop *should* be ..."?

extensions/ext/GL_EXT_control_flow_attributes2.txt
```diff
@@ -0,0 +1,123 @@
...
+    and /C/ in *peel_count(C)*/*partial_count(C)* must be a 32-bit positive signed
+    or unsigned non-specialization compile-time constant integral expression, or a
+    compile-time error results.
+
+    In addition, /M/ in *iteration_multiple(M)* must be greater than 0.
+
+
+Issues
+
+    1.
```
gnl21, line 117: If there are no issues then we can drop this section altogether.

gnl21 (commented):
A couple of editorial suggestions. The only substantive comment is clarifying whether 0 is allowed for the decorations.

extensions/ext/GL_EXT_control_flow_attributes2.txt
wooyoungqcom: I think M must be > 0. So, I dropped the sentence "In addition, ...". 
Do you think it is OK?

extensions/ext/GL_EXT_control_flow_attributes2.txt
wooyoungqcom: done

extensions/ext/GL_EXT_control_flow_attributes2.txt
wooyoungqcom: done

extensions/ext/GL_EXT_control_flow_attributes2.txt
gnl21: So we're now saying that `I`, `M` and `C` must all be > 0? It is at least possible to have min_interations of 0, although there's not much point in adding a hint for that. I'm not too clear on whether a peel_count or partial_count of 0 makes sense as a hint.

Although I think it's well defined that "positive" mean "> 0", a quick survey of the people sitting near me suggests that using it this way is not too clear. If "> 0" is correct then I'm OK to merge this, but I think that maybe:
 saying:
```suggestion
    and /C/ in *peel_count(C)*/*partial_count(C)* must be a 32-bit signed
    or unsigned non-specialization compile-time constant integral expression greater than 0, or a
```
would be clearer.

wooyoungqcom: Thank you so much !!!
That reads much clearer. I updated the extension spec.

PR #224, opened 2023-12-12
Update GL_KHR_vulkan_glsl for Rect textures

gnl21:
These are not used in Vulkan because the API does not distinguish these images as different dimensions, allowing mixing-and-matching with samplers that do and do not use unnormalized coordinates.

This update removes the additional Rect handles that were added but could never be used (texture2DRect, etc), explicitly reserves the tokens that already exist in GLSL (sampler2DRect, etc), and adds a brief explanation of how to get this same functionality in Vulkan.

Fixes Vulkan-Docs issue #1582.

### DIFF
--- a/extensions/khr/GL_KHR_vulkan_glsl.txt
+++ b/extensions/khr/GL_KHR_vulkan_glsl.txt
@@ -33,8 +33,8 @@ Status
 
 Version
 
-    Last Modified Date: 16-Mar-2020
-    Revision: 48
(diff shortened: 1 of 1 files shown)


gnl21:
Approved by the Vulkan WG 2024/01/31. The Vulkan spec does not support the "SampledRect" capability which would be required to use "Rect" images, so there shouldn't be any problem on the Vulkan side,

PR #245, opened 2024-08-25
Extend grammar tokens with their GLSL representation.

dj2:
Currently the grammar section of the spec lists all of the tokens returned by the lexer. It can be difficult to match those lexical tokens back to the relevant GLSL strings. This CL changes the token list to be a table, providing the lexical token along with the GLSL which the token represents.

This makes it easier to map things like `HIGH_PRECISION` to `highp` when reading the grammar.

The float, integer and boolean lexical tokens are linked back to the relevant portion of the spec which defines those values.

The `TYPE_NAME` and `FIELD_SELECTION` entries are mapped to `identifier` as that seems to be what they refer too.


chapters/grammar.adoc
```diff
@@ -9,84 +9,242 @@ The tokens returned from lexical analysis are
...
+| |
+| LEFT_OP | `<<`
+| RIGHT_OP | `>>`
+| INC_OP | `++`
+| DEC_OP | `--`
+| LE_OP | `\<=`
+| GE_OP | `>=`
+| EQ_OP | `==`
+| NE_OP | `!=`
+| AND_OP | `&`
```
dj2, line 203: I just realized, these three are, possibly, wrong. I think they should actually be `&&`, `^^` and `||` ?

gnl21: Yes, these ought to be the logical version, not the bitwise one.

dj2: Done.

PR #247, opened 2024-08-25
Simplify some grammar constructs.

dj2:
A number of constructs in the grammar were only a single entry, or only used once (or both). This CL inlines many of these single entry constructs to the place of usage.

This makes it simpler to read the grammar requiring less jumping around to determine what is permitted.

### DIFF
--- a/chapters/grammar.adoc
+++ b/chapters/grammar.adoc
@@ -100,11 +100,9 @@ endif::GLSL[]
 
 [role="bnf"]
 --
-_variable_identifier_ : ::
-    _IDENTIFIER_
 
 _primary_expression_ : ::
-    _variable_identifier_ +
+    _IDENTIFIER_ +
     _INTCONSTANT_ +
     _UINTCONSTANT_ +
     _FLOATCONSTANT_ +
@@ -116,7 +114,7 @@ endif::GLSL[]
 
 _postfix_expression_ : ::
     _primary_expression_ +
-    _postfix_expression_ _LEFT_BRACKET_ _integer_expression_ _RIGHT_BRACKET_ +
+    _postfix_expression_ _LEFT_BRACKET_ _expression_ _RIGHT_BRACKET_ +
     _function_call_ +
     _postfix_expression_ _DOT_ _FIELD_SELECTION_ +
     _postfix_expression_ _INC_OP_ +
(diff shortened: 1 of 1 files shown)


chapters/grammar.adoc
```diff
@@ -171,6 +160,9 @@ _postfix_expression_.
 endif::ESSL[]
 ====
 
+_function_definition_ : ::
```
dj2, line 163: This one I just moved up, it was weirdly hanging out at the very end of the grammar, so I put it with some other function definitions.

gnl21:
I made quite a lot of simplifying changes to the grammar a few years ago but they were rejected by the working group at the time, who didn't want to make updates to the grammar at all. I'd be very happy to see the grammar improved, because I think the current version is quite a mess, but I'd like to hear that the working group will accept changes here and to see how these changes integrate with my old changes before we merge this.

gnl21:
Sure, I'd be happy to push the changes. I'll dig them out (it's been a while). I agree about the jumping around and unneeded entries. A lot of it feels like it came from trying to manage data inside a yacc file rather than being a clear description of the language.

My changes also didn't change the language that the grammar accepted, which I think is good while the grammar is in its current state. I do think that having the grammar accept an over-broad language and then restricting it in text is a mistake though, so it might be interesting to look at that in the future.

chapters/grammar.adoc
```diff
@@ -297,9 +286,6 @@ _identifier_list_ : ::
     _COMMA_ _IDENTIFIER_ +
     _identifier_list_ _COMMA_ _IDENTIFIER_
 
-_function_prototype_ : ::
-    _function_declarator_ _RIGHT_PAREN_
-
```
gnl21, line 302: I've made a different change to remove this annoying rule in #252. We do need to keep _function_prototype_ (it's referred to in _function_definition_, above), but if we structure it properly then this step isn't needed at all.

dj2: Ah, good catch, I'd missed that one earlier because `function_definition` was weirdly at the end of the grammar instead of with the other functions.

dj2: I merged it into `function_defintion` to keep this Cl correct, but either way these rules are very confusing to read.

arcady-lunarg:
The bison grammar in glslang follows the grammar in the GLSL spec pretty directly, so it would be nice to see an accompanying glslang change to update that grammar as well.

gnl21:
I agree in general that it's nice to keep the glslang grammar in sync, if only because the new grammar should be simpler than the old one, but I don't think it should be a requirement for simplifying the documentation. We should keep the spec grammar as simple and accurate as possible because it's a powerful tool for saying what's valid, while glslang may want to differ from that for implementation reasons. (Either to simplify the code or potentially to generate better error messages where "Syntax error" would be unhelpful).

PR #251, opened 2024-09-11
Simplify grammar for identifier_list

gnl21:
An identifier list was defined as one or more identifiers preceded by a comma, so an additional rule was needed for a single identifier. Change the identifier_list to be what would normally be expected, a comma-separated list of one or more identifiers, and remove the additional rule.


gnl21:
@dj2 this was my first grammar change that was not taken up. The grammar is full of things like this that I consider to be bizarre and unhelpful definitions of commonplace things.

dj2 (approved):

gnl21:
Update to glslang to use the simpler grammar here: https://github.com/KhronosGroup/glslang/pull/3772

PR #252, opened 2024-09-11
Simplify grammar for function headers

gnl21:
This makes the grammar rules match how we normally think about function definitions, which makes them easier to understand.


dj2 (approved):

pdaniell-nv (approved):
OpenGL/ES WG approved

PR #253, opened 2024-09-11
Simplify grammar for init-declarators

gnl21:
Use a single grammar term for declarators (identifier followed by an optional array-specifier) and init-declarators (declarators with optional initialisation). This simplifies a few rules and makes things a bit more consistent.


gnl21:
@dj2 These 3 changes are the simplifications that I ended up implementing in my compiler. There are definitely more things that would make the grammar simpler, and interestingly our changes don't really intersect all the much, so we can probably do both. If there's interest in maintaining the grammar then there are further things that I'd like to fix.

dj2 (approved):

pdaniell-nv (approved):
OpenGL/ES WG approved

PR #254, opened 2024-09-14
Update syntax highlighting

dj2:
This Cl switches to the `rouge` highlighter which matches the Vulkan spec. The `rouge-extended-css.rb` file is copied in from the Vulkan-Docs repo.

The code examples are updated to use `source,glsl` instead of `source,c++`. This means when mousing over the code blocks in the HTML version the syntax will correctly be identified as GLSL.


gnl21:
Consistency with Vulkan is good and having the code identified correctly in the HTML is also good. Thanks for doing this.

When I build with this branch, though, the highlighting doesn't look as good as in the Vulkan spec and not quite as good as without this change (too many bold typefaces and bright colours). I'll try to track down the difference and we can merge this. Do you get consistency with the Vulkan spec when you build from this branch?

dj2:
There are two possiblities I can think of with the highlighting that I need to look into. 1. the colouring for GLSL is just different from C++, so it looks different 2. I missed some custom CSS code that Vulkan is providing and need to copy over.

Does the vulkan spec contain GLSL code, or just c/c++?

gnl21:
There is some GLSL in the Vulkan spec. I was looking, for example, at the first page of the "Fixed-Function Vertex Processing" chapter (or search for "GLSL example"). The spec source does identify it as glsl.

The 3rd possibility is that there's just something messed up with my toolchain or my build machine, since I've not tried comparing the rendering between the specs before. I was comparing the PDF versions as well, which may be different from the HTML, just because they're easier to view in my environment.

dj2:
Interesting, comparing the HTML rendering for the layout code from `Fixed-Function Vertex Processing` to a simplar chunk in `Layout Qualifiers`, they're almost identical. Comparing the two in the PDF they're very different (colours switch to green and bolding issues).

I'm guessing I'm missing some configuration for the PDF renderer, I'm going through the vulkan-docs repo to try to figure out what I'm missing.

dj2:
I'm not seeing anything, maybe @oddhack would know what I'm missing? For context, the generated PDF seems to get different highlighting style then the GLSL in the Vulkan spec. The HTML looks very similar, but I can't see why the rendered PDF is different.

oddhack:
The source highlighter is a plugin for asciidoctor. I can't say without investigation whether asciidoctor-pdf is using the same one as asciidoctor HTML output, but in any event, in Vulkan there's an extension that modifies the rouge CSS at runtime, config/rouge-extend-css.rb in Vulkan-Docs, to replace the 'github' theme color for accessibility reasons. If the GLSL source-highlighter is set to rouge and this extension is included, hopefully code will render very similarly to the Vulkan spec.

Ideally we would just put the rouge CSS changes into our own CSS, but the internal details of how rouge plugs into asciidoctor mean that does not work, so injecting them via the extension is needed (short of getting rouge upstream to adopt our changes).

That said, even if asciidoctor-pdf is using rouge, I'm not sure these CSS changes would be relevant to the PDF output. The PDF renderer goes down a completely different rendering path and does not use the HTML CSS - it has its own, extremely limited styling system. See config/themes/pdf-theme.yml in Vulkan-Docs, none of which bears on the source highlighter.

oddhack:
> The source highlighter is a plugin for asciidoctor. I can't say without investigation whether asciidoctor-pdf is using the same one as asciidoctor HTML output, but in any event, in Vulkan there's an extension that modifies the rouge CSS at runtime, config/rouge-extend-css.rb in Vulkan-Docs, to replace the 'github' theme color for accessibility reasons. If the GLSL source-highlighter is set to rouge and this extension is included, hopefully code will render very similarly to the Vulkan spec.
> 
> Ideally we would just put the rouge CSS changes into our own CSS, but the internal details of how rouge plugs into asciidoctor mean that does not work, so injecting them via the extension is needed (short of getting rouge upstream to adopt our changes).
> 
> That said, even if asciidoctor-pdf is using rouge, I'm not sure these CSS changes would be relevant to the PDF output. The PDF renderer goes down a completely different rendering path and does not use the HTML CSS - it has its own, extremely limited styling system. See config/themes/pdf-theme.yml in Vulkan-Docs, none of which bears on the source highlighter.

Edit: OK, having actually read the PR, I see you are already doing everything I suggested. So I have no actually helpful advice - getting into the guts of the browser inspector tool to figure out what styles are being applied might be the only way to figure it out.

oddhack:
> I'm guessing I'm missing some configuration for the PDF renderer, I'm going through the vulkan-docs repo to try to figure out what I'm missing.

The PDF theme for Vulkan is under config/themes/pdf-theme.yml. It has only very modest changes from the default asciidoctor-pdf theme.

dj2:
I pulled in more bits from vulkan-docs but they don't seem to have made a difference. I'm unable to build the PDF to compare what it looks like locally.

Looking at the PDFS locally, before this change:

Screenshot 2024-10-11 at 19 08 32

After this change:

Screenshot 2024-10-11 at 19 09 37

Even if it doesn't match the Vulkan spec, I think the new version looks a bit better. The highlighting is more consistent with GLSL keywords then what it picks up as C.

oddhack:
> I pulled in more bits from vulkan-docs but they don't seem to have made a difference. I'm unable to build the PDF to compare what it looks like locally.

If you can run docker or podman, you can use the same image that our gitlab CI uses for building the specs - see the 'image:' line in .gitlab-ci.yml for the image name. I should set this up on github, too.

core.adoc
```diff
@@ -56,7 +56,7 @@ endif::ESSL[]
 :numbered:
 :imagewidth: 800
 :fullimagewidth: width="800"
-:source-highlighter: coderay
+:source-highlighter: rouge
```
gnl21, line 59: Looks like this is the missing piece that is stopping it matching Vulkan:
```suggestion
:source-highlighter: rouge
:rouge-style: github
```
If this doesn't break the HTML rendering for you then I think we can make this update and merge the PR.

dj2: Done, and thank you. Apparently it's the most obvious things that I can't see, heh.

PR #268, opened 2024-11-21
Move GLSL_KHR_cooperative_matrix to GL_KHR_...

gnl21:
The extension name as given in the document is GL_KHR_cooperative_matrix so update the filename to match.


gnl21:
@jeffbolznv I'm not sure why there's a mix of these GLSL and GL extension names, but I'd like to clean them up if I can. Do you know if the difference here was intentional?

jeffbolznv:
When we started doing GLSL extensions for Vulkan several extension documents got named with the GLSL prefix, to distinguish them from GLSL-for-OpenGL. I don't have a strong opinion on how the files are named.

gnl21:
As things have turned out I think this is now making things less clear rather than more, so I'm going to go ahead with the renaming unless people shout. The extension names themselves don't match the filenames (and we show the extension name in most places), and the convention hasn't been applied consistently (eg. GL_KHR_vulkan_glsl).

jeffbolznv:
I'm OK with the change. Are you also going to update the links in the Vulkan spec?

PR #275, opened 2025-01-29
Add GLSL_EXT_nontemporal_keyword

HildarTheDorf:
The nontemporal image/memory operand in SPIR-V is not exposed in GLSL. This extension adds it.
I believe this was mentioned by @jebscape inside Khronos earlier today.

Questions:

Should this keyword be treated like restrict and implicitly stripped when calling a function, or like volatile/coherent where it is an error?

Should this be in the EXT namespace or elsewhere?


JebScape:
“Should this keyword be treated like restrict and implicitly stripped when calling a function, or like volatile/coherent where it is an error?”

Treat it similar to volatile and coherent. Do not silently strip as this could cause unintentional behavior.

JebScape:
There is an inconsistency with some of the replaced language. Let’s have them all use the following:

“restrict, coherent, volatile, and nontemporal memory”

JebScape:
@arcady-lunarg

JebScape:
The current implementation that was merged into GLSLANG requires the Vulkan memory model. Maybe that shouldn’t be the case, but I can live with it. In any case, this spec extension doesn’t mention anything about that restriction.

arcady-lunarg:
@gnl21 What's the procedure for merging extension PRs?

extensions/ext/GLSL_EXT_nontemporal_keyword.txt
```diff
@@ -0,0 +1,162 @@
...
+      and nontemporal may be used in the declaration of buffer variables
+      (i.e., members of shader storage blocks).
+
+    Replace the sentence beginning "When calling user-defined functions,
+    opaque-type variables qualified with coherent, volatile, readonly,
+    writeonly may not be passed to functions whose formal parameters lack such
+    qualifiers." with the following:
+
+      When calling user-defined functions, opaque-type variables qualified with
+      coherent, volatile, readonly, writeonly, and nontemporal may not be passed
```
gnl21, line 106:
```suggestion
      coherent, volatile, readonly, writeonly, or nontemporal may not be passed
```
or you could use the phrasing "any combination of ... " which is used elsewhere.

gnl21:
I think this should probably use the EXT namespace, in which case it can be merged whenever the authors are happy with it. If you want to use the global namespace then that will need to be approved by the working group.

I've made one wording suggestion and it would be nice to get the restrictions consistent between the spec and glslang before we merge.

extensions/ext/GLSL_EXT_nontemporal_keyword.txt
```diff
@@ -0,0 +1,162 @@
...
+                  again in the near future, and should not be cached
+
+    After the paragraph beginning "Memory accesses to image variables
+    declared using the restrict qualifier may be compiled..." add the
+    following paragraph:
+
+      Memory accesses to image variables declared using the `nontemporal`
+      qualifier may be compiled to assume their value will not be needed
+      again in the near future. This allows the compiler to avoid bringing
+      that data unnecessarily into cache.
```
gnl21, line 90: Should probably add here "It is a compile-time error to use the `nontemporal` qualifier when not generating SPIR-V"

JebScape: Not sure I’d do this. Better to just leave it as a hint. There’s nothing wrong with other backend compilers deciding to use this.

extensions/ext/GLSL_EXT_nontemporal_keyword.txt
JebScape: The original language used “and,” so it doesn’t make sense to change that here. If you want to change the base spec, then that’s outside the scope of this extension.

JebScape:
> I think this should probably use the EXT namespace, in which case it can be merged whenever the authors are happy with it. If you want to use the global namespace then that will need to be approved by the working group.
> 
> I've made one wording suggestion and it would be nice to get the restrictions consistent between the spec and glslang before we merge.

One point that was mentioned to me when I brought up to Hildar the lack of Vulkan memory model being specified was that another extension that required it also did not make any mention of it.

I haven’t pressed the issue because I honestly don’t think it *ought* to be limited to just the Vulkan memory model, but since this would involve significantly more work to GLSLANG to build in support for GLSL450, it’s probably not likely to be done right now.

I’d prefer it to get done eventually rather than constrain the spec to only be limited to the Vulkan memory model.

Instead, as a stopgap, this was merged in: https://github.com/KhronosGroup/glslang/pull/3874

I get the feeling that Hildar is done putting significant work into this. It was volunteer work in the first place. I’m also out of time to put much into this myself. I’m inclined to just leave things as they are right now and merge things as is.

extensions/ext/GLSL_EXT_nontemporal_keyword.txt
gnl21: The quotation above ("Replace the sentence...") doesn't include any conjunction here and the referenced version of the spec seems to say "or".

JebScape: I just looked at 4.10 Memory Qualifiers in the spec and it uses “and”
IMG_5131

gnl21:
I'm reluctant to merge spec extensions that say something is going to work when it doesn't because I don't think this makes for a good experience for developers who come to use this later. I guess since this isn't guaranteeing anything in the first place it's fine to implement it by just ignoring it.

JebScape:
The GL_KHR_memory_scope_semantics requires the Vulkan memory model but doesn’t specify it, setting the precedent here. That said, I would not oppose an effort to update both to have it mentioned. Entirely up to you.

extensions/ext/GLSL_EXT_nontemporal_keyword.txt
JebScape: Oh I see now. Yes that should be corrected.

JebScape: The quote is also incorrect and missing the full language.

gnl21:
Thanks. GL_KHR_memory_scope_semantics is essentially the extension for using the Vulkan memory model in GLSL and I don't think it will ever be supported without the memory model. We should probably make that clearer in the extension spec. I'm fine with glslang#3874 for this extension.

HildarTheDorf:
@JebScape @gnl21 Rebased and restored the original wording in 8.11

JebScape:
I am satisfied with the latest changes. All good on your end @gnl21?

gnl21:
Yes, looks good. Sorry for the delay.

PR #277, opened 2025-02-16
GLSL_EXT_buffer_reference with in/out interface

spencer-lunarg:
Discussed a lot in https://gitlab.khronos.org/spirv/SPIR-V/-/issues/779 this is untested and there is a `VUID-StandaloneSpirv-Input-09557` in Vulkan and this should not be allowed, so explicitly marking it

cc @jeffbolznv as author and also @gnl21

### DIFF
--- a/extensions/ext/GLSL_EXT_buffer_reference.txt
+++ b/extensions/ext/GLSL_EXT_buffer_reference.txt
@@ -18,8 +18,8 @@ Status
 
 Version
 
-    Last Modified Date:         October 25, 2024
-    Revision:                   3
+    Last Modified Date:         February 16, 2025
+    Revision:                   4
 
 Number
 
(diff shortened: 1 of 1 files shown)


gnl21:
Sorry, my fault.

Reading the spec change, the thing that it sounds like it's trying to ban is already banned in the text, and I can't construct something I think this change should apply to. Can you give an example of some shader code that you're talking about?

spencer-lunarg:
```glsl
#version 450
#extension GL_EXT_buffer_reference : enable

layout(buffer_reference) buffer X {
    int payload;
};
layout(location = 1) flat in X x;

void main() {}
```

hmm, I guess re-reading this, it sound like I ment `layout(location = 1) flat in buffer_reference X x;` which ya, is already banned

If you think the current GLSL spec is "good as is" then feel free to close. If you think I am just messing up the GLSL terms, happy to update the PR too

Issue #280, opened 2025-04-02
Allow/add OpenGL only extension GL_EXT_shader_realtime_clock with GLES


ashleysmithcol:
The extension [EXT_shader_realtime_clock](https://github.com/KhronosGroup/GLSL/blob/main/extensions/ext/GL_EXT_shader_realtime_clock.txt) is only available in OpenGL due to the specification wording:

    Dependencies

    This extension is written against Revision 8 of the version 4.40 of the
    OpenGL Shading Language Specification, dated June 16, 2014.

    This extension builds upon features introduced by the
    GL_ARB_shader_clock extension.

    This extension interacts with GL_ARB_gpu_shader_int64.

It should be possible to port this extension to be available with GLES. As this is the only GL_ extension apart from GL_ARB_gpu_shader_int64 that was not written against or allows GLES, there aren't any prior examples of porting to GLES. One option is to add wording to the EXT_shader_realtime_clock extension to allow in GLES.

Tests do not exist in piglit this specific extension but should be relatively easy to create if we copy ARB_shader_clock tests: https://gitlab.freedesktop.org/mesa/piglit/-/tree/main/tests/spec/arb_shader_clock/execution?ref_type=heads "clock.shader_test" and "clock2x32.shader_test"

Tests do exist in Vulkan for the similar extension [VK_KHR_shader_clock](https://github.com/KhronosGroup/VK-GL-CTS/blob/main/external/vulkancts/modules/vulkan/shaderexecutor/vktShaderClockTests.cpp) so support should be high.

Use cases include being able to timestamp beginning and end of shader stages to aid in performance tracking without using vendor specific tools.

If this is acceptable I will create a MR with the specification wording.

gnl21:
The linked MR was approved by the working group and has been merged, so I think this issue is fixed.

PR #281, opened 2025-04-08
Add ESSL support for GL_EXT_shader_realtime_clock

ashleysmithcol:
Discussed on issue here: https://github.com/KhronosGroup/GLSL/issues/280

### DIFF
--- a/extensions/ext/GL_EXT_shader_realtime_clock.txt
+++ b/extensions/ext/GL_EXT_shader_realtime_clock.txt
@@ -16,6 +16,7 @@ Contributors
     Daniel Rakos, AMD
     Timothy Lottes, AMD
     Graham Sellers, AMD
+    Ashley Smith, Collabora
 
 Status
 
@@ -23,8 +24,8 @@ Status
 
 Version
 
-    Last Modified Date: 10/29/2019
-    Revision: 2
+    Last Modified Date: 2025/04/07
+    Revision: 3
 
 Number
 
@@ -39,6 +40,8 @@ Dependencies
     GL_ARB_shader_clock extension.
 
     This extension interacts with GL_ARB_gpu_shader_int64.
+
+    This extension requires OpenGL 4.4 or OpenGL ES 3.1.
 
 Overview
 
@@ -83,7 +86,7 @@ Modifications to the OpenGL Shading Language Specification, Version 4.50
 Additions to Chapter 8 of the OpenGL Shading Language Specification
 (Built-in Functions)
 
-    Add New Section 8.19, "Timing Functions"
+    Add New Section 8.18, "Timing Functions"
 
     Syntax:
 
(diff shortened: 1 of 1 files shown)


ashleysmithcol:
@zmike can I ask if this is the correct location for this change? The extension is under the GLSL repo, not the OpenGL repo since it's for Vulkan, but is also available on OpenGL

zmike:
I think we can do it from public too? @gnl21

README.adoc
```diff
@@ -101,3 +101,4 @@ that do not live in the Khronos registries for OpenGL or OpenGL ES.
 - link:{repo}/nv/GLSL_NV_cluster_acceleration_structure.txt[GL_NV_cluster_acceleration_structure]
 - link:{repo}/nv/GLSL_NV_linear_swept_spheres.txt[GL_NV_linear_swept_spheres]
 - link:{repo}/ext/GLSL_EXT_nontemporal_keyword.txt[GL_EXT_nontemporal_keyword]
+- link:{repo}/ext/GL_EXT_shader_realtime_clock.txt[GL_EXT_shader_realtime_clock]
```
gnl21, line 104: This already seems to exist at line 57.

extensions/ext/GL_EXT_shader_realtime_clock.txt
```diff
@@ -40,6 +41,9 @@ Dependencies
 
     This extension interacts with GL_ARB_gpu_shader_int64.
 
+    Additional changes are written against the OpenGL ES Shading
+    Language Specification, Version 3.10.4, dated 2016/01/29.
+
```
gnl21, line 46: I think we would normally write something like:
```suggestion
    This extension requires OpenGL 4.4 or OpenGL ES 3.1.

```
(I copied this from https://registry.khronos.org/OpenGL/extensions/KHR/KHR_shader_subgroup.txt which is defined for both GL and ES.)

extensions/ext/GL_EXT_shader_realtime_clock.txt
```diff
@@ -107,6 +111,11 @@ Dependencies on GL_ARB_gpu_shader_int64:
     If GL_ARB_gpu_shader_int64 is not supported, remove clockRealtimeEXT() and any
     reference to the uint64 type.
 
+Modifications to the OpenGL ES Shading Language Specification, Version 3.10.4
+
+    Incorporate all applicable modifications to the OpenGL Shading Language
+    Specification, Version 4.40 above.
+
```
gnl21, line 118: I don't think this part is needed. If we want to call out any specific differences between the way this applies in ES and GL then we can do that here, but if there's nothing to write then we can remove the whole section.

gnl21:
Yes, this is a fine place to have the change. I'd like the ES working group to agree that it's OK for this to be extended to apply in ES before we merge it.

extensions/ext/GL_EXT_shader_realtime_clock.txt
ashleysmithcol: I believe I found this text in another extension but removing works

gnl21:
GL/ES working group approved.

PR #282, opened 2025-04-15
GL_QCOM_tile_shading

wooyoungqcom:
This extension addes support for QCOM tile shading functionaltiy in GLSL

### DIFF
--- a/extensions/qcom/GLSL_QCOM_tile_shading.txt
+++ b/extensions/qcom/GLSL_QCOM_tile_shading.txt
@@ -1,0 +1,326 @@
+Name
+
+    QCOM_tile_shading
+
+Name Strings
+
+    GL_QCOM_tile_shading
+
+Contact
+
+    Wooyoung Kim, Qualcomm Technologies, Inc. (wooykim 'at' qti.qualcomm.com)
+
+Contributors
+
+    Jeff Leger, Qualcomm
+    Rithwik Kollipara, Qualcomm
+    Ruihao Zhang, Qualcomm
+    Matthew Smith, Qualcomm
+    Wooyoung Kim, Qualcomm
+
+Status
+
+    Draft
+
+Version
+
+    Last Modified Date: 2025-04-15
+    Revision: 1
+
+Dependencies
+
+    This extension can be applied to OpenGL GLSL versions 4.60
+    (#version 460) and higher.
+
+    This extension can be applied to OpenGL ES ESSL versions 3.10
+    (#version 310) and higher.
+
+    This extension is written against the OpenGL Shading Language
+    version 4.60.8, dated Aug 12, 2023.
+
+    This extension interacts with revision 43 of the GL_KHR_vulkan_glsl
+    extension, dated October 25, 2017.
+
+Overview
+
+    This extension adds support for tile shading functionality in GLSL.
+
+    Most mobile GPUs utilize tile-based deferred rendering (TBDR) combined with 
+    high-bandwidth "tile memory" to optimize for power and performance. 
+    In a TBDR architecture, color and depth buffer attachments are partitioned
+    into a grid of small regions called "tiles" and the tile rendering passes
+    render into the tiles allocated in specialized high-bandwidth on-die memory
+    called "tile memory".  The TBDR architecture creates an opportunity to express
+    operations that execute per-tile, while the framebuffer content is still cached
+    in the tile memory. 
+
+    The extension describes new GLSL language features that allow shader writers to take
+    advantage of the functionality that the Vulkan QCOM tile shading extension offers.
+    In addition, it adds new input layout qualifiers for compute shaders to facilitate
+    area-based dispatch that proportionally maps compute threads to pixels in a tile 
+    and to determine the optimal split between workgroup sizes and workgroup counts.
+
+
+Modifications to the OpenGL Shading Language Specification, Version 4.60.8
+
+    Including the following line in a shader can be used to control the
+    language features described in this extension:
+
+      #extension GL_QCOM_tile_shading : <behavior>
+
+    where <behavior> is as specified in Section 3.3.
+
+    A new preprocessor #define is added: 
+
+      #define GL_QCOM_tile_shading 1
+
+
+    Modify Section 4.4, Layout Qualifiers (p. 65)
+
+    (add to the Layout Qualifier table, pp. 66-69)
+
+                                              Qualifier  Individual             Block     Allowed
+    Layout Qualifier                            Only      Variable      Block   Member   Interfaces
+    ----------------------------------------  ---------  -------------  -----  --------  -----------
+    non_coherent_attachment_readQCOM             X          -            -        -      fragment in
+                                                                                         (Vulkan only)
+
+    tile_attachmentQCOM                          X          X            X        -      fragment/compute
+                                                                                         uniform (Vulkan only)
+
+    shading_rate_xQCOM =                         X                                       compute in
+    shading_rate_yQCOM =                                                                 (Vulkan only)
+    shading_rate_zQCOM =
+
+
+    Modify the Fragment Shader Inputs subsection under 4.4.1 Input Layout Qualifiers (p. 76)
+
+    Fragment shaders allow the following layout qualifier of in only (not with variable declarations):
+
+    layout-qualifier-id:
+        early_fragment_tests
+        non_coherent_attachment_readQCOM
+
+    The early_fragment_tests qualifier is used to request that fragment tests be
+    performed before fragment shader execution, as described in section 15.2.4
+    "Early Fragment Tests" of the OpenGL Specification.
+
+
+    (add to the end of the Fragment Shader Inputs subsection).
+
+    The non_coherent_attachment_readQCOM qualifier requests that tile image attachment
+    reads ignore rasterization order for color/depth/stencil/input attachments.
+
+    For example, 
+     layout( non_coherent_attachment_readQCOM ) in;
+
+    Only one fragment shader (compilation unit) need declare this, though more
+    than one can. If at least one declares the non_coherent_attachment_readQCOM
+    qualifier, then the non-coherent attachment read is enabled.
+
+    See the Vulkan API documentation for details on rasterization order.
+
+
+    Modify the Compute Shader Inputs subsection under 4.4.1 Input Layout Qualifiers (p. 78)
+
+    There are no layout location qualifiers for compute shader inputs.
+
+    Layout qualifier identifiers for compute shader inputs are the workgroup size qualifiers:
+
+    layout-qualifier-id :
+        local_size_x = layout-qualifier-value
+        local_size_y = layout-qualifier-value
+        local_size_z = layout-qualifier-value
+        shading_rate_xQCOM = layout-qualifier-value
+        shading_rate_yQCOM = layout-qualifier-value
+        shading_rate_zQCOM = layout-qualifier-value
+
+
+    (insert the following three paragraphs before the sentence starting with "If multiple
+    compute shaders attached to a single program object" in the Compute Shader Inputs subsection).
+
+    The shading_rate_xQCOM, shading_rate_yQCOM, and shading_rate_zQCOM qualifiers are used
+    to declare a desired shading rate within a compute shader in the first, second,
+    and third dimension, respectively. If a shader does not specify a rate for one
+    of the dimensions, that dimension will have a rate of 1.
+    
+    The shading_rate_xQCOM and shading_rate_yQCOM qualifiers are required to be a power of 2.
+    If any of the shading rate values is greater than the maximum value supported by the 
+    implementation for that dimension, a run-time validation failure will result.
+
+    If present, the qualifiers will override the local_size_x/local_size_y/local_size_z
+    layout qualifiers and the implementation will compute the workgroup sizes.  This implies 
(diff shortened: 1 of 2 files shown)


extensions/qcom/GLSL_QCOM_tile_shading.txt
```diff
@@ -0,0 +1,326 @@
...
+    render into the tiles allocated in specialized high-bandwidth on-die memory
+    called "tile memory".  The TBDR architecture creates an opportunity to express
+    operations that execute per-tile, while the framebuffer content is still cached
+    in the tile memory.
+
+    The extension describes new GLSL language features that allow shader writers to take
+    advantage of the functionality that the Vulkan QCOM tile shading extension offers.
+    In addition, it adds new input layout qualifiers for compute shaders to facilitate
+    area-based dispatch that proportionally maps compute threads to pixels in a tile
+    and to determine the optimal split between workgroup sizes and workgroup counts.
```
gnl21, line 61: I find this sentence hard to parse. What is it that is "to determine the optimal split between workgroup sizes and counts"? Is it the layout qualifiers?

Also, a nit, I'm not sure that "proportionally" here makes this clearer. I think I know what you're intending to say, but I don't think this is clear from the text.

extensions/qcom/GLSL_QCOM_tile_shading.txt
```diff
@@ -0,0 +1,326 @@
...
+                                                                                         uniform (Vulkan only)
+
+    shading_rate_xQCOM =                         X                                       compute in
+    shading_rate_yQCOM =                                                                 (Vulkan only)
+    shading_rate_zQCOM =
+
+
+    Modify the Fragment Shader Inputs subsection under 4.4.1 Input Layout Qualifiers (p. 76)
+
+    Fragment shaders allow the following layout qualifier of in only (not with variable declarations):
```
gnl21, line 98:
```suggestion
    Fragment shaders allow the following layout qualifier on in only (not with variable declarations):
```

extensions/qcom/GLSL_QCOM_tile_shading.txt
```diff
@@ -0,0 +1,326 @@
...
+    compute shaders attached to a single program object" in the Compute Shader Inputs subsection).
+
+    The shading_rate_xQCOM, shading_rate_yQCOM, and shading_rate_zQCOM qualifiers are used
+    to declare a desired shading rate within a compute shader in the first, second,
+    and third dimension, respectively. If a shader does not specify a rate for one
+    of the dimensions, that dimension will have a rate of 1.
+    
+    The shading_rate_xQCOM and shading_rate_yQCOM qualifiers are required to be a power of 2.
+    If any of the shading rate values is greater than the maximum value supported by the 
+    implementation for that dimension, a run-time validation failure will result.
```
gnl21, line 149: You say what happens if the rate is over the maximum but only that a power of 2 is required. Is it a compile error to specify something else?

extensions/qcom/GLSL_QCOM_tile_shading.txt
```diff
@@ -0,0 +1,326 @@
...
+    If present, the qualifiers will override the local_size_x/local_size_y/local_size_z
+    layout qualifiers and the implementation will compute the workgroup sizes.  This implies 
+    that uses of local_size_id_x/local_size_id_y/local_size_id_z layout qualifiers are ignored.
+
+
+    If multiple compute shaders attached to a single program object declare a fixed workgroup size,
+    the declarations must be identical; otherwise a link-time error results.  
+
+    Furthermore, if a program object contains any compute shaders, at least one must contain an input
+    layout qualifier specifying a fixed workgroup size for the program, or a link-time error will occur.
```
gnl21, line 160: Is it still true that you have to declare a fixed workgroup size if you're using the shading_rate versions of the qualifiers?

extensions/qcom/GLSL_QCOM_tile_shading.txt
```diff
@@ -0,0 +1,326 @@
...
+    and third dimension, respectively. If a shader does not specify a rate for one
+    of the dimensions, that dimension will have a rate of 1.
+    
+    The shading_rate_xQCOM and shading_rate_yQCOM qualifiers are required to be a power of 2.
+    If any of the shading rate values is greater than the maximum value supported by the 
+    implementation for that dimension, a run-time validation failure will result.
+
+    If present, the qualifiers will override the local_size_x/local_size_y/local_size_z
+    layout qualifiers and the implementation will compute the workgroup sizes.  This implies 
+    that uses of local_size_id_x/local_size_id_y/local_size_id_z layout qualifiers are ignored.
```
gnl21, line 153: Having read through this section I'm still not really sure what the qualifiers are supposed to do or what the rate means. Is it that the workgroup size is the tile size divided by the shading rate?

extensions/qcom/GLSL_QCOM_tile_shading.txt
```diff
@@ -0,0 +1,326 @@
...
+    The input variable gl_TileOffsetQCOM is filled with the framebuffer
+    coordinates of the top-left texel of the current tile.
+
+    The input variable gl_TileDimensionQCOM represents the tile dimension in pixels (x and y)
+    as well as the number of layers (z).  The gl_TileDimensionQCOM will report the same tile
+    dimensions for all tiles in a rendering pass, including "edge tiles" that may only be
+    partially rendered because the tile boundaries extend beyond the edges of the framebuffer
+    or the Vulkan API "renderArea".
+
+    The input variable gl_TileApronSizeQCOM represents the widths of an apron (in X and Y directions).
```
gnl21, line 246: I think it would be appropriate to replace this text with something that just says that they;re defined the same as in fragment shaders?

extensions/qcom/GLSL_QCOM_tile_shading.txt
matsmithdev: Correct, the new layout qualifiers for the shading rate helps the compiler/driver determine workgroup sizes and counts. Since the shading rate and tile size determine the total number of threads to be dispatched, one then needs to determine how those threads are divided into workgroups. This proportioning is implementation defined.

In this case, "proportionally" means the total number of threads is proportional to the size of the tile. E.g. a shading rate of 1x1x1 means one thread will get dispatched per pixel per layer/view.

extensions/qcom/GLSL_QCOM_tile_shading.txt
matsmithdev: The implementation defines the workgroup size and count from both the shading rate and the tile size. The shading rate defines the proportion of pixels to threads in a vkCmdDispatchTile() command. So a 1x1x1 shading rate will dispatch one thread per pixel per layer/view for each tile. A 2x2x1 rate will dispatch one thread per 2x2 block of pixels per layer/view for each tile.

extensions/qcom/GLSL_QCOM_tile_shading.txt
matsmithdev: No, the shading rate is sufficient, you don't need to then define a local size. There were discussions over this internally regarding the correct meaning of the word "fixed" here. Does this specifically mean defining a constant local size within the shader must be present, or that some layout qualifier (like the shading rate) that ultimately determines a local size must be present?

extensions/qcom/GLSL_QCOM_tile_shading.txt
wooyoungqcom: Done.

extensions/qcom/GLSL_QCOM_tile_shading.txt
wooyoungqcom: Thanks for spotting it. Fixed.

extensions/qcom/GLSL_QCOM_tile_shading.txt
gnl21:
```suggestion
    It adds new input layout qualifiers for compute shaders to facilitate
    area-based dispatch that proportionally map compute threads to pixels in a tile,
    allowing the implementation to determine the optimal split between workgroup sizes and workgroup counts.
```

extensions/qcom/GLSL_QCOM_tile_shading.txt
gnl21:
```suggestion
    The shading_rate_xQCOM and shading_rate_yQCOM qualifiers are required to be a power of 2, otherwise a compile-time error results.
    If any of the shading rate values is greater than the maximum value supported by the 
    implementation for that dimension, a run-time validation failure will result.
```

extensions/qcom/GLSL_QCOM_tile_shading.txt
gnl21:
```suggestion
    If present, the qualifiers will override the local_size_x/local_size_y/local_size_z
    layout qualifiers and the implementation will compute the workgroup sizes and counts.
    The total number of threads in each dimension will be the number of pixels in the tile
    divided by the appropriate shading_rate, but these will be distributed between workgroups
    in an implementation-defined manner.
```

I've removed the part about ignoring the local_size qualifiers because this always causes us problems in Vulkan trying to determine what this means when the ignored declarations would otherwise be invalid. We shouldn't be ignoring the error checks on these, applications should either get them right or not include them at all.

extensions/qcom/GLSL_QCOM_tile_shading.txt
gnl21: There's never been a difference between those ideas before, so we've never resolved which one it is. I would say, from a plain reading, that "fixed" applies to the local_size qualifiers and not to the shading_rate qualifiers. (The local_size qualifiers say that they "declare a fixed workgroup size", while the new paragraphs don't say that, suggesting that they don't).

Ideally I think we'd modify the first paragraph here to say specifically that these qualifiers determine a workgroup size by specifying a shading rate, then remove the word "fixed" from these 2 paragraphs to make it clear that they are intended to include the fixed and non-fixed declarations.

wooyoungqcom: We state shading_rate_*  "override" local_size_*.  Do we need to repeat the "fixed" part ?

I interpret "fixed" as the value being the same across multiple compute shaders in a single program object. What you said seems to imply that my interpretation is not exactly correct.

What is the correct interpretation or definition of "fixed" as used in the GLSL spec?  Depending on the definition, we may need to further revise the paragraph.

extensions/qcom/GLSL_QCOM_tile_shading.txt
wooyoungqcom: Fixed.

extensions/qcom/GLSL_QCOM_tile_shading.txt
wooyoungqcom: Done

extensions/qcom/GLSL_QCOM_tile_shading.txt
gnl21: I don't think a definition has ever been nailed down because when it was written there was only 1 way to specify the size, so it was obviously talking about that. We should either define it now or remove it from the spec. I find it hard to say that a local_size that changes as the tile size changes is "fixed".

The current spec includes the text
> If multiple compute shaders attached to a single program object declare
a fixed workgroup size, the declarations must be identical; otherwise a link-time error results.

which suggests that it is possible for multiple shaders to declare a "fixed workgroup size" without then all being the same, so I don't think your interpretation can be quite right.

If there is any reason to use the word "fixed" here then I think it is this:
> If the fixed workgroup size of the shader in any dimension is less than or equal to zero [ ... ], a compile-time error
results.

where "fixed" would mean "constant that the compiler can check".

Since we're discussing this anyway, do you need this extension to allow having both sets of qualifiers but ignoring one? Could it instead be made an error to ask for both a specific "local_size" and a "shading_rate"?

extensions/qcom/GLSL_QCOM_tile_shading.txt
gnl21: Yes, I think it's fine. I hadn't noticed the switch from `local_size_x` to `local_size_id_x` in the removed text.

extensions/qcom/GLSL_QCOM_tile_shading.txt
wooyoungqcom: "I find it hard to say that a local_size that changes as the tile size changes is "fixed"."

I get your point, but I think with the presence of local_size_id_*, I cannot see local_size_* are fixed, either, if the definition of "fixed" means "not changing".


If multiple compute shaders attached to a single program object declare
**a fixed workgroup size,** the declarations must be identical; otherwise a link-time error results.

What it has is "a fixed workgroup size", and I read it as the multiple compute shaders declare **the same value** but may be with different declarations.

" Since we're discussing this anyway, do you need this extension to allow having both sets of qualifiers but ignoring one? Could it instead be made an error to ask for both a specific "local_size" and a "shading_rate"? "

Yes, we would like developers to add shading_rate_x* to existing compute shaders, instead of requiring them remove all the instances of local_size_* / local_size_id_*

wooyoungqcom: Is the "fixed" used to say the workgroup size does not change throughout the shader that it is specified ? The spec says "if no local_size_* are specified, the default of (1,1,1) is assumed. If that is the case, having the paragraphs for tile_shading_rate_* seems OK.

PR #284, opened 2025-06-04
Move `Description` sections up to the top of sections.

dj2:
In the few places where a `Description` block was added due to layout the old table formatting, the `Description` is now folded into the header content for that section.


dj2:
ping

PR #287, opened 2025-06-16
Add GL_ARM_tensors

svenvh:


janharaldfredriksen-arm:
@gnl21 - this should be ready to merge. Could you have a look when you get a chance?

extensions/arm/GL_ARM_tensors.txt
```diff
@@ -0,0 +1,318 @@
...
+        void tensorWriteARM(tensorARM t, uint coords[], uint8_t data[], uint tensorOperands = 0U, ...)
+        void tensorWriteARM(tensorARM t, uint coords[], uint16_t data[], uint tensorOperands = 0U, ...)
+        void tensorWriteARM(tensorARM t, uint coords[], uint32_t data[], uint tensorOperands = 0U, ...)
+        void tensorWriteARM(tensorARM t, uint coords[], uint64_t data[], uint tensorOperands = 0U, ...)
+        void tensorWriteARM(tensorARM t, uint coords[], float16_t data[], uint tensorOperands = 0U, ...)
+        void tensorWriteARM(tensorARM t, uint coords[], float32_t data[], uint tensorOperands = 0U, ...)
+        void tensorWriteARM(tensorARM t, uint coords[], float64_t data[], uint tensorOperands = 0U, ...)
+
+    Description: Write one or more elements to Tensor *t*. The number and type of
+    the elements written are given by the parameter *data*.
```
gnl21: For reads you say that the type of elements read are given by the type of the tensor, but here by the type of the data. Is there intended to be a difference between the two, because I'm not sure what it would be. Are any conversions between the read/write types and the tensor types allowed?

extensions/arm/GL_ARM_tensors.txt
```diff
@@ -0,0 +1,318 @@
...
+    dimension of the tensor.
+    It is a compile-time error if the number of supplied coordinates does not
+    equal the rank of the tensor.
+    *tensorOperands* is a constant bitmask of zero or more gl_TensorOperands*
+    values.
+    Depending on the specified tensorOperands, additional arguments may have to
+    be specified.
+    The behavior when one or more tensor elements read are outside of the tensor
+    bounds and gl_TensorOperandsOutOfBoundsValueARM is not provided, is specified
+    by the client API.
```
gnl21: If you're reading more than 1 element do they all have to be contained within a single dimension, or can the coordinates wrap around in some way? I could imagine that if you read to the end of one row maybe it would carry on in the next.

PR #295, opened 2025-12-14
Add GL_EXT_relaxed_block_layout

Karpov-Ruslan:
This extension is written to support `VK_KHR_relaxed_block_layout` **Vulkan** extension. Currently there is only `VK_EXT_scalar_block_layout` support - `GL_EXT_scalar_block_layout`. To correct this historical injustice, I present this extension.

If GLSL Working Group approves this extension, then I will create a PR in glslang ([link](https://github.com/Karpov-Ruslan/glslang/tree/GL_EXT_relaxed_block_layout) to fully working fork with the extension implementation).

The example of usage `GL_EXT_relaxed_block_layout`:
```
#version 460 core

#extension GL_EXT_relaxed_block_layout : require

struct S {
    int i;  // Struct member offset: 0
    vec3 j; // Struct member offset: 4
};

layout(binding = 0, std430, relaxed) buffer SSBO {
    int a; // Offset: 0
    S s;   // Offset: 16 (the base alignment of structure equals to 16)
    int b; // Offset: 32 (the size of structure equals to 16)
};

void main() {...}
```

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -105,3 +105,4 @@ that do not live in the Khronos registries for OpenGL or OpenGL ES.
 - link:{repo}/ext/GL_EXT_float8_e5m2_e4m3.txt[GL_EXT_float8_e5m2_e4m3]
 - link:{repo}/ext/GL_EXT_shader_64bit_indexing.txt[GL_EXT_shader_64bit_indexing]
 - link:{repo}/ext/GLSL_EXT_shader_invocation_reorder.txt[GL_EXT_shader_invocation_reorder]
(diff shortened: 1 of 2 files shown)


Karpov-Ruslan:
@gnl21 Could you take a look, please? If I've broken the established procedure for adding a new extension or if I don't have the rights to create an extension, please let me know.

jeffbolznv:
It's not clear to me what the goal of this extension is. I think scalar block layout is supported basically everywhere at this point, so why not just use that?

PR #299, opened 2026-01-21
Add GLSL_EXT_structured_descriptor_heap

Tobski:
This extension is a work in progress, provided for comment.

https://github.com/KhronosGroup/GLSL/pull/298 provides a basic interface for interacting with descriptor heaps; this extension is intended to enable applications to interact with descriptor heaps in a similar way to uniform buffers, such that descriptor data and POD can be mixed freely within the heaps.


spencer-lunarg:
ping - is there anyone assigned to this, we are having massive issues in validation due to the current GLSL limitations

spencer-lunarg:
> Can you elaborate on what it is you want it for?

[The current GLSL to use untyped pointer is not a joy to use](https://gitlab.khronos.org/vulkan/vulkan/-/issues/4724)

We just ran into yet-another issue yesterday because we are trying to hack everything around alignments from being at the heap starting address... (so basically the lack of `heap_offset` is hard to use untyped pointers currently for anything meaningful outside small tests)

SaschaWillems:
So looking at this (thanks for bringing this up recently @spencer-lunarg ), what I'm currently trying to do:

```glsl
layout(descriptor_heap) uniform UBO {
	mat4 projection;
	mat4 view;
	mat4 model;
} ubo;
```

Doesn't seem to be possible without an extension like this. Above code actually throws an error when compiling with glslang.

But e.g. this syntax compile just fine in current GLSL:

```glsl
layout(descriptor_heap) uniform UBO {
	mat4 projection;
	mat4 view;
	mat4 model;
} ubo[2];
```

But outright doesn't work, without any mention as to why. But looking at this extension I think I understand what's going on and feels like the current way of using heaps with untyped pointers is downright useless for uniform buffers.

If so, that's prob. the reason I've been having all those issues with the untyped pointer way of accessing descriptor heaps. Would be a real bummer tbh.

SaschaWillems:
@Tobski : Any ETA on this? This has been causing me quite some trouble and it looks like other developers are starting to run into similar issues - not being able to use structured buffers with descriptor heaps. See e.g. https://github.com/KhronosGroup/glslang/issues/4224

With the F2F discussion it would be worthwhile to make this available soon(ish),

Tobski:
> @Tobski : Any ETA on this? This has been causing me quite some trouble and it looks like other developers are starting to run into similar issues - not being able to use structured buffers with descriptor heaps. See e.g. https://github.com/KhronosGroup/glslang/issues/4224

> With the F2F discussion it would be worthwhile to make this available soon(ish),

No ETA just yet but it's high on my todo list, it's just unfortunately a long todo list. Lots of Post-F2F follow up actions I need to churn through. In theory I just have to go through and figure out some of the spec details; then we should have someone available to implement, but it'll depend on their availability too. But we do want to get this done.

Tobski:
@SaschaWillems just to check this is what you're actually looking for...

> So looking at this (thanks for bringing this up recently @spencer-lunarg ), what I'm currently trying to do:
> 
> ```glsl
> layout(descriptor_heap) uniform UBO {
> 	mat4 projection;
> 	mat4 view;
> 	mat4 model;
> } ubo;
> ```
> 
> Doesn't seem to be possible without an extension like this. Above code actually throws an error when compiling with glslang.

So I'm looking at the syntax above, and I'm guessing that you're trying to access a uniform buffer at offset zero in the heap? In GL_EXT_descriptor_heap we only allowed for arrays, which is why only the second of your examples works. If you want that to work as-is, that should probably be a separate bug against the existing extension. 

For this extension, expressing that, in the current extension, would be like:

```glsl
layout(heap_offset = 0) resourceheap {
    uniform {
         mat4 projection;
         mat4 view;
         mat4 model;
    } ubo;
} myHeap;
```

That syntax is unlikely to actually work as-is because of GLSL grammar, so it's probably to be something like:

```glsl
layout(buffer_type) uniform MyUbo {
     mat4 projection;
     mat4 view;
     mat4 model;
};

layout(heap_offset = 0) resourceheap MyHeap {
    MyUbo ubo;
} myHeap;
```

Is this actually what you're after?

SaschaWillems:
Looking at my comment above I wasn't 100% clear. Sorry for that.

So what I'm after is this:

```glsl
layout (set = 0, binding = 0) uniform UBO {
	mat4 projection;
	mat4 view;
	mat4 model[2];
} ubo[2];
``` 

Which is what I'm using in an actual descriptor heap sample that uses descriptor set mappings instead of untyped pointers:.

This compiles just fine with the descriptor heap modifier:

```glsl
layout(descriptor_heap) uniform UBO {
	mat4 projection;
	mat4 view;
	mat4 model[2];
} ubo[2];
```

But does not work.

Tobski:
@SaschaWillems if what you're compiling is not working that is a separate issue - that's supposed to work and access two UBOs at the bottom of the heap. This extension is very much separate. Have you got an issue filed against glslang we can look at? If not, could you file one?

Tobski:
I've just realised you mentioned https://github.com/KhronosGroup/glslang/issues/4224 which seems to cover it

extensions/ext/GLSL_EXT_structured_descriptor_heap.txt
```diff
@@ -0,0 +1,423 @@
...
+    This does add some complexity to compilation, as various sizes and
+    alignments must now be calculated when the shader is compiled in the
+    client API.
+    
+    For example, the `MaterialData materials[];` declaration would map to the
+    following SPIR-V:
+    
+    ```spir-v
+    // Offset and Stride decorations
+                                  OpMemberDecorateIdEXT %materials_t OffsetIdEXT %offset_0 0
```
spencer-lunarg:
```suggestion
                                  OpMemberDecorateIdEXT %materials_t 0 OffsetIdEXT %offset_0
```

arcady-lunarg:
@Tobski What's the status of this? Is it still WIP or is it more or less finalized now and ready to be merged?

Tobski:
I'm waiting for feedback to settle down on https://github.com/KhronosGroup/glslang/pull/4272 - if we're confident that this is good to go, we can merge this spec.

Tobski:
Looking at the glslang PR I think we're good, so I've resolved conflicts and it's no longer WIP. This can be merged.

Tobski:
Oh wait _I_ can merge it. ....I'll just do it I guess???

arcady-lunarg (approved):

PR #317, opened 2026-03-23
Add EXT_abort

r-potter:
cc @gnl21

### DIFF
--- a/extensions/ext/GLSL_EXT_abort.txt
+++ b/extensions/ext/GLSL_EXT_abort.txt
@@ -1,0 +1,136 @@
+Name
+
+    EXT_abort
+
+Name Strings
+
+    GL_EXT_abort
+
+Contact
+
+    Tobias Hector, AMD (tobias.hector 'at' amd.com)
+
+Contributors
+
+    Contributors to the GL_EXT_debug_printf extension
+
+Status
+
+    Draft
+
+Version
+
+    Last Modified Date:         September 31, 2024
+    Revision:                   1
+
+Number
+
+    TBD
+
+Dependencies
+
+    This extension requires GL_KHR_vulkan_glsl
+
+Overview
+
+    This extension adds an "abortEXT" built-in function that takes a string
+    message and a list of arguments, similar to debugPrintfEXT.
+
+Modifications to GL_KHR_vulkan_glsl
+
+    Add to the "Mapping to SPIR-V" section
+
+      abortEXT -> OpAbortKHR
+      literal string -> OpConstantDataKHR, 8-bit uint array type with UTFCodePoints decoration
+
+Modifications to the OpenGL Shading Language Specification, Version 4.60.8
+
+    Including the following line in a shader can be used to control the
+    language features described in this extension:
+
+        #extension GL_EXT_abort : <behavior>
+
+    where <behavior> is as specified in section 3.3
+
+    New preprocessor #defines are added to the OpenGL Shading Language:
+
+        #define GL_EXT_abort     1
+
+    Modify section 3.1 Character Set and Phases of Compilation 
+
+    Add to the list of supported characters:
+
+        single quote ('), double quote ("), backslash (\)
+
+
+    Replace the paragraph beginning with "There are no digraphs or trigraphs"
+    with:
+    
+    There are no digraphs or trigraphs. Outside of string literals (described
+    below), there are no escape sequences or uses of the backslash beyond use
+    as the line-continuation character
+
+
+    Replace the paragraph "There are no character or string data types, so
+    no quoting characters are included" with:
+
+    There are no character or string data types, but literal strings can be
+    used. A literal string is an initial double quote character, followed by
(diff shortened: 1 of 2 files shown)


arcady-lunarg:
Can this be merged? It's already implemented in glslang.

rdb:
One grammar question: like `GL_EXT_debug_printf`, this extension doesn't actually define a syntactic path to allow a literal string in a function-call argument. The wording ("the first argument must be a literal string") suggests the string is a special argument rather than an expression; yet glslang parses strings as a primary-expression, so `abortEXT(("foo"))` is accepted by glslang even though nothing in this extension suggests that it should be.

So: what is actually intended here? I would like to know so that so that the grammar I have proposed in `GL_EXT_spirv_intrinsics_string` (#333) can be kept consistent with what `GL_EXT_abort` and `GL_EXT_debug_printf` do. Adding an explicit production to the extension text may be useful to remove any ambiguity.

Also, minor typo: "Last Modified Date" mentions "September 31, 2024", even though September has only 30 days.

jeffbolznv:
I think you're right that there is a missing edit to the grammar, that `("foo")` should not be treated as a literal string, and that glslang is missing validation for this for debugPrintf (and presumably also for abort, but I haven't looked).

extensions/ext/GLSL_EXT_abort.txt
```diff
@@ -0,0 +1,136 @@
...
+
+    The abort string may include format specifiers for further processing,
+    though it is implementation-defined whether this processing occurs.
+    Applications may process the stream themselves.
+        
+    If an implementation does process format specifiers, additional arguments
+    after the string must match the type indicated by the order of the format
+    specifiers in the string. If there are insufficient arguments for the
+    format specifiers, the behavior is undefined. If the format is exhausted
+    while arguments remain, the excess arguments are ignored.
```
r-potter:
```suggestion
    format specifiers, the behavior is undefined. If the format is exhausted
    while arguments remain, the excess arguments will still be consumed by the
    implementation.
```

HansKristian-Work:
Wasn't this okay to merge now?

r-potter:
It should be, yes. It was blocked on the literal string question, but I implemented that for both debug printf and abort

rdb:
There are no grammar productions listed yet, is that intentional? As-is there's still no grammar rule to enable passing a literal string as an argument to a function.

rdb:
I would suggest something like this. If that is considered unnecessary if the existence of such a rule can be inferred from the prose, then never mind.

```
Additions to Chapter 9 of the OpenGL Shading Language Specification
(Shading Language Grammar)

    Add the following rule:

        function_call_argument:
            assignment_expression
            STRING_LITERAL

    Replace the "function_call_header_with_parameters" rule as follows:

        function_call_header_with_parameters:
            function_call_header function_call_argument
            function_call_header_with_parameters COMMA function_call_argument
```

r-potter:
I would have no objections to doing this, although if we do then we should probably also do so for printf. 

@gnl21 @jeffbolznv Do you have opinions on this one?

gnl21:
I agree that it would good to add something like @rdb suggests. I was expecting to see something like that in response to Jeff's comment.

gnl21:
Since we now have 3 extensions that need the same thing, maybe we should update the core grammar with this so that all the extensions can refer to it. I think (but I'd have to check) that this wouldn't be a problem for the unextended spec if STRING_LITERAL were unconditionally banned.

extensions/ext/GLSL_EXT_abort.txt
```diff
@@ -0,0 +1,137 @@
...
+        
+    If an implementation does process format specifiers, additional arguments
+    after the string must match the type indicated by the order of the format
+    specifiers in the string. If there are insufficient arguments for the
+    format specifiers, the behavior is undefined. If the format is exhausted
+    while arguments remain, the excess arguments will still be consumed by the
+    implementation.
+    The set of format specifiers is implementation-dependent, but must
+    include at least "%d" and "%i" (int), "%u" (uint), and "%f" (float).
+
```
r-potter, line 133: Leaving this here for now, although I like @gnl21's suggestion of a core grammar rule as an alternative:

```suggestion
    Add the following grammar rule to chapter 9:

        function_call_argument:
            assignment_expression
            STRING_LITERAL

    Replace the "function_call_header_with_parameters" rule as follows:

        function_call_header_with_parameters:
            function_call_header function_call_argument
            function_call_header_with_parameters COMMA function_call_argument
            
```

r-potter:
> Since we now have 3 extensions that need the same thing, maybe we should update the core grammar with this so that all the extensions can refer to it. I think (but I'd have to check) that this wouldn't be a problem for the unextended spec if STRING_LITERAL were unconditionally banned.

Would the proposal here be:

* Add the change to the core grammar
* Explicitly ban it in 6.1 Function Definitions so that user defined functions can't do the same thing

Happy to make the change, just need to know exactly what we want

PR #321, opened 2026-04-22
Add GLSL extensions component (glslext)

oddhack:
Per https://github.com/KhronosGroup/Vulkan-Site/issues/169


oddhack:
@gnl21 I'd like to go ahead and merge this immediately if that's OK with you. It does not affect the independent spec PDF / HTML builds, just adds new targets to generate an Antora component and import the refpages into it.

gnl21:
Sure, looks good to me. I'll merge it now.

PR #330, opened 2026-06-10
Update status from Draft to Complete in GLSL_EXT_spirv_intrinsics

Tobski:


rdb:
Would it be possible to defer this until I have proposed my amendments? I will try to get to it soon.

Tobski:
This is functionally not a draft - we have a lot of code relying on the existing extension in production. That it's not tagged as such is spec bug, it's not reflective of reality. Any new functionality needs to be a new extension IMO.

rdb:
That's fine by me.

I have discovered a potential omission in the GL_EXT_spirv_intrinsics text. It doesn't specify whether spirv_literal can be used with a `set`-qualified instruction (ie. OpExtInst). The current glslang implementation doesn't seem to allow this (there's an assertion error). That's defensible, but then there should probably be wording in the spec to disallow this explicitly.

Tobski:
> That's fine by me.

Cool. I'll merge this then.

> I have discovered a potential omission in the GL_EXT_spirv_intrinsics text. It doesn't specify whether spirv_literal can be used with a `set`-qualified instruction (ie. OpExtInst). The current glslang implementation doesn't seem to allow this (there's an assertion error). That's defensible, but then there should probably be wording in the spec to disallow this explicitly.

That is actually supposed to be allowed... do you have an example shader for this? And could you raise a glslang issue for it?

rdb:
Done, see KhronosGroup/glslang#4317

Tobski:
Thanks!

Issue #331, opened 2026-06-13
Extending GL_EXT_spirv_intrinsics


rdb:
`GL_EXT_spirv_intrinsics` maps GLSL function declarations to SPIR-V instructions, but a `spirv_instruction` prototype must be fixed-arity, and its arguments are limited to 32-bit int/bool/float literals or by-id objects. Two classes of instruction therefore can't be defined in a header today:

* instructions with a variable-length operand tail, and
* instructions taking literal-string operands.

An example of such an instruction that is established today is `debugPrintfEXT`, which cannot currently be expressed using `GL_EXT_spirv_intrinsics`. As discussed in KhronosGroup/glslang#2679, I would like to add this ability, extending the usefulness of this extension to both existing and potentially new use cases.

Based on the fact that `GL_EXT_spirv_intrinsics` is established and ([de facto](https://github.com/KhronosGroup/GLSL/pull/330)) no longer in Draft, I understand I should propose new extensions. I could fold it into a single extension but the functionality cleanly separates along these two orthogonal axes:

### Variadic tail support (`GL_EXT_spirv_intrinsics_variadic`)

This would permit a `spirv_instruction`-qualified function declaration to end in a trailing `...`, which allows 0..n arguments to be passed in. Default encoding is by id, matching the default. For completeness, `spirv_literal ...` could be made valid which would encode the whole tail as literal operands (a la OpVectorShuffle):

```glsl
// OpVectorShuffle (literal tail)
spirv_instruction(id = 79)
vec4 shuffle(vec4 a, vec4 b, spirv_literal ...);
```

### String support (`GL_EXT_spirv_intrinsics_string`)

This would add a new `spirv_string` pseudo-type, which can only occur in a `spirv_instruction`-qualified function declaration. It would also allow literal strings to be passed to variadic arguments. They would normally be encoded as an OpString id.

An open question is the combination `spirv_literal spirv_string`. Current NonSemantic extensions take an OpString by id, but literal string operands are well-established in SPIR-V, such as in OpEntryPoint, where the characters are encoded in successive uint32_t arguments until a null byte is encountered. As far as I know there is no current use case for this, as I'm not aware of any such instruction that can appear in a function body, but it does kind of fall naturally out of the specification so I would be willing to add it for completeness' sake.

Using `spirv_by_reference spirv_string` would not be allowed since no pointers to string literals can be constructed.

The combination of these extensions allows fully and cleanly capturing `GL_EXT_debug_printf`:

```glsl
[...]

PR #332, opened 2026-06-27
Add GL_EXT_spirv_intrinsics_variadic extension

rdb:
This is the first half of #331, adding the extension text for the variadics part (the forthcoming string extension is designed to be independent). I've also got a working implementation in glslang, which I will submit separately.

The extension adds variadic syntax (like `void f(int x, ...)`) for `spirv_instruction` functions, motivated by use cases like `debugPrintfEXT` and `abortEXT`. The grammar is written to restrict the use of the ellipsis to only `spirv_instruction`-qualified functions, though I expect most implementations will reject invalid uses semantically rather than purely at the grammar level.

I had underestimated how much work there was to define variadic functions properly, especially regarding overload resolution, which in practice won't come up much but must be addressed. The crux of it is this:

GLSL short-circuits function resolution: if there's an exact type match between the formal parameters and the calling arguments, that match is used and overload resolution is pre-empted. I've kept that rule intact by excluding variadic declarations from ever being considered an "exact match", so any call involving a variadic always goes through normal overload resolution. As it happens, glslang's existing variadics support (used only for builtins) already behaves this way.

Where I deliberately deviate from the current glslang behavior is that glslang currently reports `f(float)` and `f(float, ...)` as ambiguous when called with `f(1)` but chooses the former silently with `f(1.0)` as per the exact-match rule, which I deemed too surprising. I have resolved this by adding a tie-break rule making it prefer the non-variadic overload in this specific case. The alternatives all turned out messy; the cleanest would require effectively redefining function calling in GLSL by gutting the exact-match rule, which I deemed to be far too invasive for an extension of this kind. Issue 5 lays out the arguments in more detail; the chosen option does have a pleasing analogy in C++.

Another point of note is that I made this against 4.60.8 since I could not locate a copy of the 4.60.7 spec that the base extension is based on. I can submit a PR to update the base extension or verify my extension against a copy of the 4.60.7 spec if someone can point me to one, if this is considered a problem.

This is my first time submitting an extension, so please let me know if I have made mistakes. Thanks for your consideration!

### DIFF
--- a/extensions/ext/GLSL_EXT_spirv_intrinsics_variadic.txt
+++ b/extensions/ext/GLSL_EXT_spirv_intrinsics_variadic.txt
@@ -1,0 +1,308 @@
+Name
+
+    EXT_spirv_intrinsics_variadic
+
+Name Strings
+
+    GL_EXT_spirv_intrinsics_variadic
+
+Contact
+
+    Reinier de Blois (me 'at' rdb.name)
+
+Contributors
+
+    Reinier de Blois
+    Tobias Hector, AMD
+
+Status
+
+    Draft
+
+Version
+
+    Revision: 1
+    Last Modified Date: 2026-06-24
+
+Dependencies
+
+    This extension is written against version 4.60.8 of the
+    OpenGL Shading Language Specification, dated August 14, 2023.
+
+    This extension requires GL_EXT_spirv_intrinsics.
+
+Overview
+
+    GL_EXT_spirv_intrinsics allows a SPIR-V instruction to be exposed to
+    GLSL by declaring a function whose signature matches the instruction's
+    operands. That mechanism requires the declared function to be of fixed
+    arity: it has exactly as many parameters as the instruction has operands.
+
+    A number of SPIR-V instructions, however, end in a variable-length tail
+    of operands. A notable example is the DebugPrintf instruction in the
+    NonSemantic.DebugPrintf extended instruction set. Such instructions cannot
+    be expressed with GL_EXT_spirv_intrinsics today.
+
+    This extension allows the use of a trailing ellipsis (`...`) as the last
+    entry in the parameter list of a function declaration qualified with
+    `spirv_instruction`, permitting a variable number of trailing arguments to
+    be supplied at the call site and appended to the instruction's operands.
+
+    As with GL_EXT_spirv_intrinsics generally, the compiler cannot validate
+    that the resulting operand sequence is correct for the target
+    instruction. It is strongly recommended that generated SPIR-V be validated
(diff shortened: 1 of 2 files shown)


extensions/ext/GLSL_EXT_spirv_intrinsics_variadic.txt
```diff
@@ -0,0 +1,308 @@
...
+
+        #version 460
+
+        #extension GL_EXT_spirv_intrinsics : enable
+        #extension GL_EXT_spirv_intrinsics_variadic : enable
+        #extension GL_EXT_spirv_intrinsics_string : enable
+
+        spirv_instruction(extensions = ["SPV_KHR_non_semantic_info"],
+                          set = "NonSemantic.DebugPrintf", id = 1)
+        void debugPrintfEXT(spirv_string format, ...);
```
Tobski, line 207: This example using spirv_string is weird, given that's not introduced here. I suggest picking something else as the example for this spec so it can be a standalone.

extensions/ext/GLSL_EXT_spirv_intrinsics_variadic.txt
```diff
@@ -0,0 +1,308 @@
...
+       `spirv_by_reference`?
+
+    PROPOSED: Yes.
+
+    A by-pointer operand tail falls out naturally and is symmetric with the
+    `spirv_literal` and default (by-id) cases. It is included for completeness.
+
+    2) Should the ellipsis be permitted on user-defined functions?
+
+    PROPOSED: No.
```
Tobski, line 226: This should be stated somewhere in the actual spec body - I don't see anything banning this. Issues are only for clarifications, but they should be backed up by the spec.

Tobski (changes requested):
Couple comments but overall looks good to me - should have some more eyes before accepting though, and ideally an implementation in at last draft form. Thanks for putting this together!

extensions/ext/GLSL_EXT_spirv_intrinsics_variadic.txt
rdb: This is the main grammar production added:
```
spirv_instruction-qualifier function_prototype_with_variadic_tail SEMICOLON
```
There's no reference to the variadic tail anywhere else, so a variadic tail on user-defined functions is already invalid syntax - should there be explicit prose banning it as well?

extensions/ext/GLSL_EXT_spirv_intrinsics_variadic.txt
rdb: You're right, I'll make up some example that works without `spirv_string`.

PR #333, opened 2026-06-28
Add GL_EXT_spirv_intrinsics_string extension

rdb:
This is the second half of #331, adding the string part (see #332 for the variadic extension). They are separate extensions because they describe orthogonal features with minimal interaction. The glslang implementation will follow soon.

This extension builds on `GL_EXT_spirv_intrinsics` and adds a `spirv_string` pseudo-type that is only valid as a formal parameter type in a `spirv_instruction`-qualified function declaration. It furthermore provides the ability for literal string arguments to be passed to those parameters, as well as to variadic tails (when the variadic extension is enabled). It is lowered to an OpString whose result id is used as the operand.

The main conundrum I've had to work through was determining whether literal strings are treated as primary expressions (which matches what glslang implements, eg. allowing a parenthesized `("foo")`), or can only be used as function call arguments (which matches what can actually be derived from the motivating `GL_EXT_debug_printf` extension, though it is a bit light on the actual grammar).

I have deliberately decided to stick with the most restrictive reading of what is actually specified, since doing otherwise would ask this extension to do much more work to more generally define strings as a type and expression. I deemed that to be out of scope for an extension that just aims to represent functions like `debugPrintfEXT` as a `spirv_instruction`. It may be worth revisiting this in a future extension that reusably and more generally defines string expressions in GLSL, or the referenced extensions should be revised to define the grammar in more detail and this one changed to match.

### DIFF
--- a/extensions/ext/GLSL_EXT_spirv_intrinsics_string.txt
+++ b/extensions/ext/GLSL_EXT_spirv_intrinsics_string.txt
@@ -1,0 +1,333 @@
+Name
+
+    EXT_spirv_intrinsics_string
+
+Name Strings
+
+    GL_EXT_spirv_intrinsics_string
+
+Contact
+
+    Reinier de Blois (me 'at' rdb.name)
+
+Contributors
+
+    Reinier de Blois
+
+Status
+
+    Draft
+
+Version
+
+    Revision: 1
+    Last Modified Date: 2026-06-27
+
+Dependencies
+
+    This extension is written against version 4.60.8 of the
+    OpenGL Shading Language Specification, dated August 14, 2023.
+
+    This extension requires GL_EXT_spirv_intrinsics.
+
+    This extension interacts with GL_EXT_spirv_intrinsics_variadic.
+
+    This extension interacts with GL_KHR_vulkan_glsl.
+
+Overview
+
+    GL_EXT_spirv_intrinsics allows a SPIR-V instruction to be exposed to
+    GLSL by declaring a function whose signature matches the instruction's
+    operands. It also defines the lexical machinery for literal strings.
+    However, it does not add syntax to pass a string to a function parameter,
+    so a SPIR-V instruction that takes an OpString <id> as operand cannot be
+    expressed with GL_EXT_spirv_intrinsics today.
+
+    To that end this extension adds a `spirv_string` pseudo-type, which may be
+    used as the type of a parameter in a function declaration qualified with
+    `spirv_instruction`. A literal string supplied for such a parameter is
+    turned into an OpString, and that OpString's <id> is passed as the operand.
+    When GL_EXT_spirv_intrinsics_variadic is also enabled, a literal string may
+    likewise be supplied as a trailing variadic argument.
+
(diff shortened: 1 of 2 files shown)


extensions/ext/GLSL_EXT_spirv_intrinsics_string.txt
```diff
@@ -0,0 +1,333 @@
...
+        A literal string may be empty (""), which produces an OpString whose
+        operand contains only the terminating NUL character.
+
+        If GL_EXT_spirv_intrinsics_variadic is enabled, a literal string
+        argument may also match an unqualified variadic tail of a function
+        declaration qualified with `spirv_instruction`. In that case, an
+        OpString is also generated and its result <id> is appended as the
+        corresponding operand. It is a compile-time error for a literal string
+        to match a variadic tail with a `spirv_by_reference` or `spirv_literal`
+        qualifier.
```
Tobski, line 142: I would suggest that the tail has to be qualified with spirv_string to make this work - non-string arguments ignore it, but strings will be converted to OpString arguments.

The rationale for this is that otherwise you are making concrete, with no get out, how strings should be passed to intrinsics, and I'd like to leave the door open to passing them as byte arrays in future.

By requiring spirv_string here, it leaves the door open for strings-as-byte-arrays to be passed as arguments in future.

extensions/ext/GLSL_EXT_spirv_intrinsics_string.txt
```diff
@@ -0,0 +1,333 @@
...
+
+        spirv_instruction(extensions = ["SPV_KHR_non_semantic_info"],
+                          set = "NonSemantic.DebugPrintf", id = 1)
+        void debugPrintfEXT(spirv_string format, ...);
+
+        void main()
+        {
+            float x = 1.0;
+            debugPrintfEXT("x = %f\n", x);
+        }
```
Tobski, line 204: I think it would be illustrative to have an example with a string passed in the variadic tail.

Tobski (changes requested):
This looks good too - just a couple review comments and would request some more eyes and a draft impl before accepting.

extensions/ext/GLSL_EXT_spirv_intrinsics_string.txt
Tobski: I guess spirv_string isn't a qualifier per say... so maybe not exactly that? But I do think there should be something to distinguish this behavior.

rdb: Oof, that's a good point... `spirv_string` isn't a qualifier as you say (and it would probably create parser ambiguity to introduce it as one here), but some way to say "if there is a string, encode it as OpString" sounds like it would be necessary... I will think about this.

extensions/ext/GLSL_EXT_spirv_intrinsics_string.txt
rdb: Sure. I could modify the example to add an "%s" argument to printf.

PR #335, opened 2026-07-04
Add GL_EXT_function_control_attributes

Karpov-Ruslan:
This extension allows to use `[[inline]]` and `[[noinline]]` attributes on functions to give a hint to the compiler.

One of the side effects of this extension is the ability to fine-tune function inlining when using *spirv-opt*. Let's say I have a GLSL compute shader:
```glsl
#version 460

layout(set = 0, binding = 0, std430) buffer SSBO {
    uint data;
};

void func() {
    // Imagine that there is quite a lot of code here.
    atomicAdd(data, 1U);
    atomicAdd(data, 2U);
    atomicAdd(data, 3U);
    atomicAdd(data, 4U);
    atomicAdd(data, 5U);
    atomicAdd(data, 6U);
    atomicAdd(data, 7U);
    atomicAdd(data, 8U);
    atomicAdd(data, 9U);
}

void main() {
    // There are a lot of `func` calls here.
    func();
    func();
    func();
    func();
    func();
    func();
    func();
    func();
    func();
    func();
}
```

I'll take this shader and compile it to SPIR-V using *glslang* (https://godbolt.org/z/adfrb5soW), and then I'll want to use *spirv-opt* with the `-Os` flag, which should reduce the size of incoming SPIR-V. But in this case the size will be increased (https://godbolt.org/z/fxv5rjGo9). However, by adding the `[[noinline]]` attribute to the `func` function, the `DontInline` Function Control bit will be added to this function in SPIR-V, so *spirv-opt* will not inline this function and the size of the outgoing SPIR-V will indeed be reduced (https://godbolt.org/z/d6daKEK5n).

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -114,3 +114,4 @@ that do not live in the Khronos registries for OpenGL or OpenGL ES.
 - link:{repo}/nv/GLSL_NV_explicit_typecast.txt[GL_NV_explicit_typecast]
 - link:{repo}/nv/GLSL_NV_cooperative_matrix_decode_vector.txt[GLSL_NV_cooperative_matrix_decode_vector]
 - link:{repo}/ext/GL_EXT_ocp_microscaling_types.txt[GL_EXT_ocp_microscaling_types]
+- link:{repo}/ext/GL_EXT_function_control_attributes.txt[GL_EXT_function_control_attributes]
--- a/extensions/ext/GL_EXT_function_control_attributes.txt
+++ b/extensions/ext/GL_EXT_function_control_attributes.txt
@@ -1,0 +1,137 @@
+Name
+
+  EXT_function_control_attributes
+
+Name Strings
+
+  GL_EXT_function_control_attributes
+
+Contact
+
+  Ruslan Karpov
+
+Contributors
+
+  Ruslan Karpov, independent
+
+Status
+
+  Draft.
+
+Version
+
+  Last Modified Date:   July 04, 2026
+  Revision: 1
+
+Dependencies
+
+  This extension can be applied to OpenGL GLSL versions 1.40
+  (#version 140) and higher.
+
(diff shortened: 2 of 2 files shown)


spencer-lunarg:
tl;dr - agree that GLSL should just have a simple way to just expose this and :100: support this idea

@Karpov-Ruslan I realize that the goal is to stop `spirv-opt`, but what is the goal after that

Basically I have tried doing this for GPU-AV inside VVL and found that NO ONE actually will honor your `DontInline` function control mask... the worst part is there is no way to query for support of it neither

`NVIDIA 553.31+` does support it, but we have found periodic issues with it, for example, we have to always turn it off for Ray Tracing shaders

... so more of a "heads up"... if you only care about SPIR-V binary size, that is a VERY valid reason as well!

Karpov-Ruslan:
> Basically I have tried doing this for GPU-AV inside VVL and found that NO ONE actually will honor your DontInline function control mask... the worst part is there is no way to query for support of it neither

To be honest, I didn’t even consider it. Sad to hear that. However, SPIR-V Function Control is about hints to the compiler. If the driver developers don't take this into account, it's their full right. Perhaps in the future more and more drivers will support this, so we need to be prepared.

Karpov-Ruslan:
I slightly changed the specification so that `inline` and `noinline` could only be used in attributes.

Karpov-Ruslan:
Since the implementation was merged, I think this PR can be merged too.

Karpov-Ruslan:
@gnl21 Is there anything I need to do on my end? It would be nice to merge the spec for this extension before the new Vulkan SDK is released.

gnl21:
No, sorry, this is fine to merge. There are conflicts with the base branch now, so could you either resolve the conflicts or give me permission to modify the branch and I can do it myself?

Karpov-Ruslan:
> give me permission to modify the branch and I can do it myself

Done

PR #339, opened 2026-08-05
Add GL_EXT_optional_input_attachment_index

### LINKED ISSUE #338: Making input_attachment_index optional
As you probably know, in Vulkan the InputAttachmentIndex decoration for SubpassData images is now optional thanks to VK_KHR_dynamic_rendering_local_read. However, glslang and GL_KHR_vulkan_glsl make it mandatory. Some CTS tests that I'm writing would benefit from making the decoration optional in GLSL, because it leads to more interesting scenarios and, at the same time, writing the shaders in SPIR-V assembly would be quite painful. What's the best way to make the decoration optional for GLSL? Do you think an extension is required? Could we simply tune the wording in GL_KHR_vulkan_glsl?

Related:

* https://github.com/KhronosGroup/glslang/pull/4361
* https://github.com/KhronosGroup/Vulkan-Docs/issues/2776

rg3igalia:
This makes input attachment indices optional so they can be used as described in VK_KHR_dynamic_rendering_local_read without having to write SPIR-V assembly.

Closes #338.

### DIFF
--- a/extensions/GL_EXT_optional_input_attachment_index.txt
+++ b/extensions/GL_EXT_optional_input_attachment_index.txt
@@ -1,0 +1,98 @@
+Name
+
+    EXT_optional_input_attachment_index
+
+Name Strings
+
+    GL_EXT_optional_input_attachment_index
+
+Contact
+
+    Ricardo Garcia (rgarcia 'at' igalia.com), Igalia
+
+Notice
+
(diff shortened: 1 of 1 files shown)


zmike:
@gnl21 Who does the approval for something like this?

gnl21:
Vulkan working group has approved.

PR #342, opened 2026-09-02
Move GL_EXT_optional_input_attachment_index.txt to the right place

rg3igalia:
This MR also links the extension from the README.adoc file and marks it as complete. Sorry I didn't realize about these details before.

CC @gnl21


gnl21:
I noticed that it wasn't linked to the index just after I pressed merge and was planning to fix it after the meeting, so this is great, thanks!

